PR #2, opened 2022-08-29
feat: clerk integration

chronark:


steven-tey:
Went with next-auth instead so I'm gonna close this! Thank you! :)

PR #3, opened 2022-08-29
fix: retry on collission and wrap in try-catch

chronark:


steven-tey:
@chronark the code evolved a lot since this PR – would you be able to update it? Thank you! 🙏

PR #10, opened 2022-09-22
Fix visual bug on stats page

MatthewStanciu:
When you open `/stats/[slug]` directly instead of in a modal, it looks a bit broken:

Screen Shot 2022-09-22 at 12 18 18 PM

This PR inherits the color instead of setting it to gray, and adds some padding, so that it looks good both in and out of the modal.

Screen Shot 2022-09-22 at 12 19 37 PM

It might actually be better to remove the gray altogether if the page isn't in a modal 🤔

### DIFF
--- a/components/stats/index.tsx
+++ b/components/stats/index.tsx
@@ -37,7 +37,7 @@ export default function Stats({
   );
 
   return (
-    <div className="py-20 mx-auto lg:px-0 px-2.5">
+    <div className="bg-gray-50 py-20 mx-auto lg:px-0 px-2.5">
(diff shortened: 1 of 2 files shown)


steven-tey:
Ahh crap, I just saw this – I've actually fixed it by changing the bg for that page's bg to `bg-gray-50`

The issue with this PR is since I'm reusing this component in multiple parts, it breaks the view for the modal as well 😅

CleanShot 2022-09-22 at 13 49 31

MatthewStanciu:
Ah whoops, glad you fixed it!

PR #13, opened 2022-09-22
Optimize WebGL and canvas

shuding:


steven-tey (approved):
Just tested this out in localhost and it LGTM (my Mac's fans are not spinning anymore)! Thank you so much for the PR @shuding!

PR #14, opened 2022-09-22
fix: some URLs not being parsed on the homepage

rwlc:
Fixes an error with some URLs being parsed incorrectly. See:

image


steven-tey (approved):
Great catch! And thank you so much for the PR! 🤩 🙏

PR #18, opened 2022-09-23
Fix typo in README

Nutlope:
Little typo that shows an extra parentheses after the deploy button (check current README). I just removed it


steven-tey:
Dangit I merged another identical commit (#24) right before I saw this, sorry @Nutlope 😅

PR #21, opened 2022-09-23
change to public

ImBIOS:
It is not a private project so I change the private to false


steven-tey:
Good catch! Also added license info, thanks!

PR #43, opened 2022-09-27
fix: consistent spacing in dashboard page in mobile view

harel-tussi:


Mr-Wesson (approved):

steven-tey (approved):
Reverted the height change but the padding inconsistency fix makes sense! Thanks for the PR! :)

PR #51, opened 2022-10-01
fix: fix desktop modal not being clickable

Snazzah:
There was an issue when clicking desktop modals that clicking *in* the modal exits it. This just makes sure the target is from the actual element rather than anything inside it.


steven-tey:
Thanks a lot for this PR! I actually made a very stupid mistake (added `onClick={() => closeModal()}` to the modal itself (and not the backdrop) so it was a quick one-line fix! Appreciate your PR though so I'm still gonna merge this (even if there are no changes)

PR #65, opened 2022-10-05
fix: fix key regex pattern when managing links

Snazzah:
[This commit](https://github.com/steven-tey/dub/commit/f8b4bda4ad3d7f0cb0cbd48afd088e02273cb8df) forgot to account for numbers in links, noticed it while making other changes.


steven-tey:
Oof good catch! Merging – thank you!

PR #68, opened 2022-10-08
feat: add QR code modal

Snazzah:
Adds a QR button to link cards, allowing you to generate QR codes out of it.
This also involved adding and modifying the source code for https://github.com/zpao/qrcode.react, since I needed to use some of it's internal functions to generate the codes to be available to download.


steven-tey:
Damn this is brilliant, great work! 😄

Gonna take a look at this tomorrow and merge it!

steven-tey:
This shows up on the canvas but we still need to figure out how to export it (the exports only have the QR code for now)

CleanShot 2022-10-11 at 23 37 48

CleanShot 2022-10-11 at 23 38 47

Snazzah:
I've updated the code so the PNG/JPEG downloads also gets the image inside the QR code. Also tried to scan the first QR code in your post and it didn't seem to work (since too many parts were taken out) so I bumped up the error level to have it work with the logo.

Should there be a toggle to get the QR code without the logo?

Sidenote: For the SVG download to work, the logo URL has to be an exact URL, so I updated that.

Snazzah:
The SVG logo is directly using the logo URL, so any localhost URL won't work if you are trying to show the SVG logo in your browser. You could add a rule to set it to Dub's URL while in development.

Also, seems like the toggle isn't working correctly, there isn't a prop that's setting the value so that may be why.

steven-tey:
hmm that doesn't seem to be it:

CleanShot 2022-10-12 at 23 21 06

Plus fetching https://dub.sh/static/logo.svg wouldn't work when downloading PNG/JPG, I suspect it's something to do with CORS:

CleanShot 2022-10-12 at 23 20 56

steven-tey:
The toggle seems to work for me too:

CleanShot 2022-10-12 at 23 22 47

Snazzah:
Ah, nevermind the toggle, I forgot to update my Tailwind config. As for the error, it might be something with CORS. Probably check the Network tab and see if that errored. I'll probably end up embedding a data URI in with the SVG export.

steven-tey:
Sorry was sleeping!

Just checked this and I actually used the `href` approach when debugging but it ended up giving me a question mark on Mac Previews (but when you expand it it looks fine):
CleanShot 2022-10-13 at 07 12 37

The CORS error is still there for PNG & JPGs though 😕 pushed a hotfix for now:

```
...
onClick={async () =>
  download(
    await getQRAsSVGDataUri({
      ...qrData,
      ...(showLogo && {
        imageSettings: {
          ...qrData.imageSettings,
          src: logo || "https://dub.sh/static/logo.svg",
        },
      }),
    }),
    "svg",
  )
}
```

steven-tey:
We're good to merge! Thank you so much again for your help with this – it's hands-down the best PR I've seen on this project so far 🙌

PR #70, opened 2022-10-08
fix: any `ua`

mxkaske:
Replacing `any` with the `ReturnType` of the stored value for more type-safety. Based on:

https://github.com/steven-tey/dub/blob/3e05edcaac0dcf8159e07174d52b21703a0cb4f8/lib/upstash.ts#L99


steven-tey:
Great catch, thank you for this! It LGTM but I'll take one last look in the AM and merge it! 😁

steven-tey:
Merged! Thank you so much again! :)

mxkaske:
Glad I could help!

PR #72, opened 2022-10-09
added focus trap on modals

msmps:
This is made possible via:

* `focus-trap-react` package which is a light weight wrapper around the popular `focus-trap` package.

I noticed there is also a focus-trap component exposed via `@headlessui/react` although that does not appear to cycle the trapped focus continuously and requires you to shift+tab to reverse whereas the above package cycles.

An additional `<div />` wrapper was required as the focus-trap expects a single child element.


steven-tey:
LGTM, merging!! Thanks a lot for thinking about a11y with this PR! 🙏

PR #87, opened 2022-10-24
💄 Added Icons to Menu Items

max-programming:
## Now Menu Items have icons
image
image

image


max-programming:
With the latest commit, I refactored the same code for the icon and text to a new `MenuIcon` component

steven-tey:
This is brilliant, thank you so much for the PR! Extra kudos to the creation of the `IconMenu` component which helps standardize the way we make icon menus :)

I'm also thinking if we should adopt https://lucide.dev/ for our icons system – it's tree-shakeable and has a ton of great options + we won't need to manually configure each icon 🤔 

Will create a discussion topic!

PR #88, opened 2022-10-24
Change placeholder on login page

max-programming:
Both login and register pages should have consistent placeholders.

### Problem
On the register page, the placeholder was `panic@theredis.co`, but on the login page, the placeholder was Email Address

I think both of them should be the same


steven-tey:
Good catch, thank you!

PR #94, opened 2022-10-31
feat: Allow links to be removed from the archive

Snazzah:
Right now, archived links still have an "Archive" button. This PR should replace that with a "Remove from archive" button that does just that.


steven-tey:
So...I think we'll probably switch to a "tags" system in the future (so `archived` will become `tag: "archived"`) so this will change then.

Merging this for now anyways, thank you so much for the PR! 🙏

steven-tey (approved):

PR #95, opened 2022-10-31
fix: fix updating redis keys when changing domains

Snazzah:
This should fix domain updates when it comes to updating Redis keys. before it was grabbing the list of keps from the earlier version of how link data was stored.


steven-tey (approved):
Great catch, thank you so much! I missed this when I migrated link relations from redis to MySQL! 

Approved & merging now! 🙏

PR #109, opened 2022-11-08
Use SMTP credentials from the environment variables

ps-89:


steven-tey:
Closing this since we recently migrated to Resend! #260

PR #125, opened 2022-11-23
Added keywords + robots

richiemcilroy:
Added keywords and robots to list of meta tags to include.


steven-tey:
Thanks for the PR Richie! I'm trying to keep the offering as minimal as possible (`title`, `description`, and `image` only) so I'm going to go ahead and close this for now – if there's demand for it in the future, we can revisit this!

PR #127, opened 2022-11-25
feat: add new implementation to fetch meta tags

rwlc:
hey @steven-tey! this pr might require some changes, as you were previously recording all the tags from the `recordMetatags` function in `@/lib/upstash`. LMK if it does need any!

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "cobe": "^0.6.2",
     "focus-trap-react": "^10.0.1",
     "framer-motion": "^7.6.12",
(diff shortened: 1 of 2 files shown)


steven-tey:
Thanks for the PR @rwlc!! Will take a look <picture data-single-emoji=":eyeses-intense:" title=":eyeses-intense:">:eyeses-intense:</picture>

steven-tey:
Will revisit https://github.com/natemoo-re/ultrahtml/issues/38 when we figure that out. Thanks a lot for the PR!!

PR #130, opened 2022-11-27
switch to node-html-parser for parsing

ynbh:
This PR makes the switch from `ultrahtml` to `node-html-parser` for parsing HTML, mainly because the former has some slight quirks in parsing weirdly-formatted attributes like `content` in meta tags. It also utilizes the library's `querySelector` API, and makes the codebase a tad bit more easy to read--and type safe--by reducing lines of code in the main logic. Using `node-html-parser` also removes the need to manually escape HTML entities that causes the JSON response to sometimes break.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,13 +41,13 @@
     "cobe": "^0.6.2",
     "focus-trap-react": "^10.0.1",
     "framer-motion": "^7.6.12",
-    "he": "^1.2.0",
     "html-escaper": "^3.0.3",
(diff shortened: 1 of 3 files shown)


ynbh:
bump @steven-tey

steven-tey (approved):
<picture data-single-emoji=":lfgtm:" title=":lfgtm:">:lfgtm:</picture> – merging! Thanks a lot for the PR! 🙏

PR #131, opened 2022-11-28
mentioned TWITTER_AUTH_TOKEN in .env.example 

theabhayprajapati:
Twitter Auth/Bearer Token is required for fetching tweets, so let's mention it in the .env.example file.


steven-tey:
Hey, thank you for the PR! I actually want to outline a list of stuff that is not required for a self-hosted version of Dub and instructions on how to remove them, and this is one of them. Closing this PR for now, but thank you for bringing it up!

PR #186, opened 2023-04-30
Fix changelog rss feed xml formatting

vahnag:
Saw your [tweet](https://twitter.com/steventey/status/1652706421610536960) and it caught my eye that you have extra commas between entries in your XML. 

This quick pull request will remove them and make the XML valid.


steven-tey:
Oooh great catch! Thank you!

PR #227, opened 2023-05-30
catch error

MenaiAla:


steven-tey:
This will be fixed as part of #246, but thanks a lot for the PR!

MenaiAla:
You are welcome.

PR #228, opened 2023-05-30
chore: upgrade to `next-auth@5`

balazsorban44:
This is a test. When NextAuth.js v5 is final, the imports will stay the same as before, but for now, the imports are from `@auth/nextjs`.

Based on https://auth-docs-git-feat-nextjs-auth-authjs.vercel.app/guides/upgrade-to-v5


steven-tey:
Taking a quick look at this now – why were there so many removed packages @balazsorban44? A lot of them don't seem related to the Auth.js upgrade

PR #249, opened 2023-07-03
fixed the property name error on Object Crisp.chat 

rashayagami:
```npm run build```  was throwing error that the property name ```onChatClose ```  on object ```Crisp.chat``` does not exists.


steven-tey:
Oh great catch! Looks like they changed `onChatClose` to `onChatClosed` in the latest version of `crisp-sdk-web` – thank you!

PR #259, opened 2023-07-09
Return 404 on link not found

Ankur-Datta-4:
Current Behaviour:
- When the link is expired or if the link is a bad one, the site shown is dub.sh landing page. Which would get users confused
https://github.com/steven-tey/dub/blob/1b0e3caf554743145f68818f8e3474050aa4c674/lib/middleware/link.ts#L66C5-L66C5

Change:
- Added a 404 return status using NextResponse in the LinkMiddleware.

Let me know if anything has to change 🙂


steven-tey:
Great callout, but will need to add a proper 404 page that adds some context about what Dub does before we can merge this PR. Something like the existing placeholder pages for domains under the free plan (https://chatg.pt/) could be a good starting point!

steven-tey:
#509 fixes this!

PR #265, opened 2023-07-13
Added search params to url parse()

yoavTaieb:
Need it for the api. route of the middleware


steven-tey:
Sorry, I'm confused – where is the `search` param being used?

Issue #263, opened 2023-07-12
Long URL breaks card layout


seanislegend:
Long URLs currently break the display of cards on the Links page. This hides the URL actions and date.

Here is an example:
<img width="720" alt="image" src="https://github.com/steven-tey/dub/assets/1477984/f356bd4d-cabd-43b7-bf47-d95719750378">

The fix requires changing the max width allowed ("fit" seems to the issue). I have used `sm:max-w-[calc(100%-10rem)]` (with 10rem being the space afforded to the right-side date and actions CTA) and it looks good. 

<img width="707" alt="image" src="https://github.com/steven-tey/dub/assets/1477984/adbf5951-a747-4a2a-9ebd-353d96a94781">

Additionally, I think moving the stats/actions to where the date is is a better layout as it affords more space to the title/URL.

<img width="711" alt="image" src="https://github.com/steven-tey/dub/assets/1477984/3735437e-01ed-4b80-a92f-b9fc972a29a0">

I have created a Tailwind replication of the issue and fixes [here](https://play.tailwindcss.com/XQjXb6Wcgn), and I'm happy to open an actual PR for the fix or my suggestion.

steven-tey:
Appreciate you surfacing this, @seanislegend, and thanks a lot for the design suggestions here – they look really good!

This is related to #261 so we definitely want to address this – let's go with the 3rd layout (your suggestion) but with a few tweaks:
1. Keep the timestamp in the same spot, but remove "Added" to save on space. 
2. Change the formatting of [timeAgo](https://github.com/steven-tey/dub/blob/main/lib/utils.ts#L141-L146) to change "ago" to a simplified version of the timestamp beyond a certain threshold (maybe 7 days?)
3. Move the QR code into the actions CTA to save on space

steven-tey:
Thanks again for the feedback + design here @seanislegend! Gave you a shoutout in the Changelog post: https://dub.sh/changelog/improved-link-cards

PR #269, opened 2023-07-14
Improve link card layout

### LINKED ISSUE #261: Short Link URL on Mobile Gets Cut Off
Mobile keeps driving me a bit nuts because I'm skimming to see if a short link exists (efficient.tips/vs/harvest), but then realize I'm skimming the full link and not the short link, so no wonder I can't find it.

The actual short link gets cut off on mobile, which is IMHO the most important URL to be able to view. Even more important than skimming the long-link.

The copy/QR/stats icons take up such valuable horizontal screen real estate.

I've made this same mistake every single time I've gone to skim for a short link now on mobile (4th time now and keep forgetting each time). So definitely feels like a rough UX on mobile in terms of hierarchy.

Wondering as well if the https:// could be scraped on the destination URL? Also not useful to visually see it in the list view. Subdomain is the first part of the destination URL that is useful to see.

steven-tey:
Fixes #261 and #263 

CleanShot 2023-07-14 at 18 06 13

Credits to @seanislegend for the link card design 🤩


seanislegend:
Great way to take the suggestion and make it better! 🫡

steven-tey:
Thanks @seanislegend! Gave you a shoutout in the Changelog post: https://dub-git-improve-link-card-elegance.vercel.app/changelog/improved-link-cards

Really appreciate the feedback!

PR #291, opened 2023-08-01
Fix `z` value on stats pages

tomheaton:
Fix `z` value on stats pages to prevent popovers displaying behind components.

Before:
image

After:
image


steven-tey:
Great catch, thank you!

PR #292, opened 2023-08-01
Prevent removing only user from project.

tomheaton:
Prevent removing the only user from a project so that it is just lost to the void.

image

Ideally, this would also need to be fixed in the UI, but this is a simple fix for now.

I have also written an email to support asking for the project back as it's just out there with no members.


steven-tey:
Modified this a bit, but thank you again for surfacing this!

CleanShot 2023-08-01 at 19 51 06

PR #310, opened 2023-08-12
Updated dicebear avatar to be the latest version

TheMikeyRoss:
I noticed the method used to fetch the avatar is not the latest as per dicebear documentation.

Plus there's no way for the user to change their avatar. Especially when their avatar does not align their race or gender. 

So I suggest fetching the cool and neutral space-invader style "Identicon" avatars using version 6.x of Dicebear.


steven-tey:
Appreciate this PR @TheMikeyRoss – we just added the ability to upload your own profile picture so I'm gonna close this for now! Thanks anyway!

CleanShot 2023-08-17 at 11 19 09@2x

PR #331, opened 2023-08-25
Add Tinybird pipes

steven-tey:
Shoutout to @sdairs for the pro-tip & help here! 🙏


sdairs:
Party time!!! 🎉

Couple things I caught:

1. There is an erroneous top level /datasources folder
2. There is a `domain_events.datasource` file in that top level /datasources folder, which isn't in the /tinybird/datasources folder
3. It seems some materialized view Pipes are missing, e.g. the `click_events_mv.datasource` exists and is created from a `pipe_for_click_events_mv` Pipe, but the `pipe_for_click_events_mv` Pipe isn't in the /tinybird/pipes dir
4. Any of the Pipes in /tinybird/pipes that are published as API Endpoint won't get published from a `tb push` at the moment. If you create a /tinybird/endpoints dir, and move the `.pipe` file that should be published as APIs into it, they will get published

steven-tey:
@sdairs Thanks for reviewing! Just realized that I can probably remove `click_events_mv` since it was created to optimize `click_events` in the first place. I just removed it and incorporated the sorting & partition key changes into `click_events` – let me know if that looks good!

Also, what would be the step-by-step instructions for someone to implement these pipes & datasource into their Tinybird instance? Do they just navigate to the `tinybird` directory and run `tb push`?

sdairs:
Looking good!

FYI any Pipe that you want published automatically as an API should move into a `/endpoints` dir, i.e., `/tinybird/endpoints` - take this structure as an example from the Web Analytics kit https://github.com/tinybirdco/web-analytics-starter-kit/tree/main/tinybird. Otherwise users will need to manually hit the publish button.

RE deployment instructions:

1. Create a Tinybird account
2. Create a new Workspace
3. Copy your Auth Token
4. In your cloned Dub repo, go to `/tinybird`
5. Install the Tinybird CLI with `pip install tinybird-cli` (requires Python >= 3.8)
6. Run `tb init` and paste your Auth Token
7. Run `tb push`
8. Voila

https://www.tinybird.co/docs/quick-start-cli.html

steven-tey:
Amazing, thank you so much @sdairs

steven-tey:
Will be continued in #491

PR #336, opened 2023-08-25
Fix default redirects not being checked on new domain

joshuaepstein:
Noticed that the default redirects weren't being checked in the middleware so fixed it.

Haven't tested yet , just submitted it.


joshuaepstein:
I've also noticed that links such as dub.co/github don't work - is this intentional

steven-tey:
Closing this since it's intentional behavior – but thank you!

PR #345, opened 2023-08-29
Create not-found.tsx

Autumn-Ou:
# default not found to replace next 404

image
image

based off the dub.co/help not found


steven-tey:
Thanks for the PR! We'll need to add a proper 404 page that adds some context about what Dub does before we can merge this PR. Something like the existing placeholder pages for domains under the free plan ([chatg.pt](https://chatg.pt/)) could be a good starting point!

PR #347, opened 2023-08-29
Add CORS header to metatags api

jparkrr:
Thanks for the project!

I was wondering if you meant to allow cross-origin requests to the metadata api? I've enabled it in this PR if you're interested.


steven-tey (approved):

steven-tey:
LGTM! Merging this now with the caveat that we might need to remove the header in the future if there are too many cases of abuse 😅

Issue #341, opened 2023-08-28
Add link to Link Inspector


fmerian:
love the new [Link Inspector](https://twitter.com/steventey/status/1695798379836796961)! 

adding a `+` to the end of the link may not be so obvious tho. 

as a user, I'd like to have a link to the Link Inspector in the footer -- under `/tools`?

![2023-08-28-hWkRorH7@2x](https://github.com/steven-tey/dub/assets/1890087/b60d97a3-79db-427f-9162-d2279b0737b2)

![2023-08-28-XzGAf2uI@2x](https://github.com/steven-tey/dub/assets/1890087/9e4627a8-d58c-42bb-856d-06638448a3e9)

cosmiclasagnadev:
@fmerian should be a quick enough task. @steven-tey would love to be assigned and work on this 🫡

steven-tey:
@cosmiclasagnadev feel free to pick this up! You should be able to take inspiration from both the current [link inspector implementation](https://github.com/steven-tey/dub/blob/main/app/inspect/%5Bdomain%5D/%5Bkey%5D/page.tsx) as well as the [Metatags API](https://github.com/steven-tey/dub/blob/main/app/dub.co/tools/metatags/page.tsx) setup to make this work! :)

steven-tey:
This shipped! <picture data-single-emoji=":ty-heart:" title=":ty-heart:"><img class="emoji" width="20" height="auto" src="https://emoji.slack-edge.com/T0CAQ00TU/ty-heart/fcc570066d2722f0.png" alt=":ty-heart:" align="absmiddle"></picture> @cosmiclasagnadev

PR #353, opened 2023-08-30
Added footer link and link inspector landing

cosmiclasagnadev:
Should address issue #341.

@steven-tey wasn't exactly sure how I would showcase the link inspector's feature on a landing page because the main feature here is "add a + to any dub link and you'll be redirecting to our link inspector first" so I opted to put an input that would redirect them to a valid link-inspect URL. Let me know if this was the wrong approach or if I should instead go and display the report button box and their link preview on the same page :D


steven-tey:
Perfect, thank you so much for the PR! Merging now :)

Issue #365, opened 2023-09-05
Undo 'Archive'


fmerian:
as a user, if I archive a link by mistake, I'd like to recover it quickly.

currently, the only option is to: 
1. show archived links then
2. unarchive link

instead, I'd like to be able to undo it.

![Undo?](https://github.com/steven-tey/dub/assets/1890087/0a5c6747-b783-41f6-97c0-d5cb5169b7e9)

RockingThor:
Can I get this issue assigned to me?

xonx4l:
Is this looks good to you @fmerian .

steven-tey:
left a comment @xonx4l 

assigned the issue to you too @RockingThor

unrenamed:
Hi @steven-tey , do you require assistance with this matter? Based on my analysis, it appears that the solution is relatively straightforward in this specific case. Let me know if I'm missing anything:

1. [Sonner's ](https://sonner.emilkowal.ski/) toast API  supports an action callback with a customizable text for the button, such as "Undo." We can use it to send the **undo** request to the API (with the opposite boolean flag in its body) whenever a user clicks the button.
2. The logic for archiving links seems unsophisticated: a boolean flag is simply modified in the database. Consequently, we can submit numerous requests to undo, redo the action, and so forth. This approach is applicable for both "Archive" and "Unarchive" state changes. Thus, it seems that only some adjustments on the UI side are necessary.

So the idea is as follows:
```ts
toast.success(`Successfully ${archive ? "archived" : "unarchived"} link`, {
    action: { label: "Undo", onClick: undoAction }
});
```

Now, the undo action can also run a "promisified" toast with the loading text for better UX:
```ts
toast.promise(sendArchiveRequest(link, !archive), {
    loading: "Undo in progress...",
    success: () => "Undo successful! Changes reverted."
});
```

However, if the aim is to make this feature adaptable to cover more app actions, the aforementioned considerations might not suffice. For instance, reversing the "Delete" action would necessitate modifications to both the database and API.

Nevertheless, if you find the solution I suggested sufficient for address this concrete "undo (un)archive" problem, I would be delighted to collaborate on implementing the feature and submitting a pull request.

steven-tey:
@unrenamed that's a fantastic analysis! The UX you suggested above sounds perfect (brilliant idea to leverage Sonner's undo API) – only suggestion that I have is maybe we can add an CMD+Z listener as well during this allow-undo period (unless the Sonner API supports this as well)

[...]

PR #368, opened 2023-09-09
FEAT: Implements the undo feature in archive .

xonx4l:
ISSUE-:#365

Implements the undo feature .


steven-tey:
@xonx4l probably can't do this since it's a serverless function (which doesn't have in-memory state). we'll need to use something like redis to make this work + we'll need to be more thoughtful on the client side design as well

PR #384, opened 2023-09-19
Update customer story

fmerian:
Added context to headlines for quick, easy read/scan of the customer story


steven-tey:
Perfect, thank you Flo!

PR #414, opened 2023-10-15
fix: Updated Readme file

abhinaisai2002:
the link for .env.example redirects to 404 page, changed to redirect to .env.example of the repo.


steven-tey (approved):
Thanks a lot!!

PR #415, opened 2023-10-15
fix: Animation of navbar background blur

chimpdev:
With this PR, I'm solving a problem by removing the backdrop blur animation. In the previous case, the transition-all class was animating everything including the backdrop blur, and this led to a strange and unpleasant animation of the backdrop blur property going from 0px to 8px. This change fixes the problem by removing the animation.

**Old**

chrome_MbEAMwPSfT

**New**

chrome_jfjyeH0MFp


chimpdev:
@steven-tey

PR #417, opened 2023-10-17
Add 404 monitoring

steven-tey:


arkhurst (approved):

beedell-roke:
@steven-tey, does anything replace this?

PR #418, opened 2023-10-18
Improve password-protected links

### LINKED ISSUE #300: Pass Password Unlock via URL Parameter
Could you pass a password via URL perameters? e.g. visiting a URL like:
https://efficient.app/apps/dub?pw=efficient

And assuming the password in this case was "efficient" it would actually enter the password and unlock it all in one fell swoop?

steven-tey:
This also closes #300


abass:
Wahhhhh! This is the absolute PERFECT timing for this 🚀🚀🚀🚀 Thank you so much @steven-tey!

PR #429, opened 2023-10-26
Fix: documantation spelling mistake

ahmod001:
I have improved the documantation by fixing a few spelling mistakes in README.md file


steven-tey:
thanks!

PR #440, opened 2023-11-02
fix settings layout sticky navbar for mobile devices

Vardhaman619:
Before :
settings page sticky navbar does not have background color and z-index , so it not look correct on mobile screen 
 
settings-page-link-does-not have-background
before-settings-page-links-overflowing

After:
added background-color and z-Index to setting-page sticky navbar  
after-fix-settings-page-sticky-navbar


steven-tey:
This has been fixed by @versecafe!

PR #473, opened 2023-11-22
fix pagination rendering 0

yuler:
image


steven-tey:
Oh yeah, great catch! Thank you!

steven-tey (approved):

PR #476, opened 2023-11-24
Add undo functionality for link (un)archiving

### LINKED ISSUE #365: Undo 'Archive'

unrenamed:
This feature resolves #365 

TODO:

- [x] add an option to undo link (un)archiving from the UI
- [x] add a logic to notify a user if an undo action is failed / succeded
- [x] discover potential ways to register a CMD+Z listener (e.g. verify if [Sonner's](https://sonner.emilkowal.ski/) API accepts custom React components)
- [x] implement a custom component / hook that triggers the undo action on button click OR when a user presses CMD (CTRL) + Z


unrenamed:
@steven-tey 👋  This PR is ready for code review.

Before you delve into the code, there are a few things to consider, and I'd love to hear your thoughts on them:

1. I've included the `<CheckCircle />` component from [Sonner's code base](https://github.com/emilkowalski/sonner/blob/main/src/assets.tsx), essentially a copy of their `<SuccessIcon />`. Personally, I'm not happy about adding another `<CheckCircleFill />` icon to your app. However, I did it to ensure a clear, consistent design. The aim was to minimize any noticeable differences between custom and default toasts. Unfortunately, Sonner's API doesn't allow rendering different custom icons based on promise status, so we're faced with a choice — either keep both icons or opt for just one. On a side note, while `<CheckCircleFill />` appears more detailed to me, if we keep both of them, there would be some visible distinctions that users might notice, especially if a default toast immediately follows our custom one.

2. Now, the good news is that I successfully added both mouse click and CMD (CTRL) + Z listeners. I introduced a new custom hook with a component that we pass to the `toast()` function, accepting custom JSX. Currently, this custom component isn't heavily customizable as I'm uncertain about potential future cases. If you have thoughts on which options we should include now (like a custom icon, undo button styles, etc.), please let me know.

3. You might also notice that I removed `await` from the calls to revalidate SWR links data. Previously, a toast only appeared after the links were reloaded, making users wait longer than necessary for any response. Without `await`, we show the toast immediately after the modal is closed, skipping the wait for the links to load. If the initial use of `await` was intentional, please inform me, and I'll revert these changes.

steven-tey:
Sorry for the delayed response here, but wow, phenomenal work @unrenamed! What a beautiful implementation – it works like a charm:

https://github.com/steven-tey/dub/assets/28986134/67880a33-6b5d-41d6-b606-eb65f6e7c175

Also really appreciate the writeup above and the amount of thought & attention to detail you put into this PR. 

1. Totally fine with adding the success icon from Sonner's codebase – I made a small edit by changing the name back to `<Success/>` instead.

2. Incredible work on adding the CMD+Z listener! I think this implementation is modular enough for now – we can iterate it in the future if needed.

3. Removing `await` in this case makes a lot of sense – thank you!

Overall, thank you so much for your help here! Will be merging this now :)

steven-tey:
@unrenamed do you have an email I can reach you at? Feel free to email me at steven@dub.co – would love to connect with you!

PR #491, opened 2023-12-10
Self-hosting Improvements

### LINKED ISSUE #12: Self host instructions
There should be instructions on how to self-host Dub.

steven-tey:
This closes #12 

image


daun-io:
Hi. Everything looks great but it seems like tinybird pipe for timeseries API is somewhat erroneous. (it seems not drawing any data on the chart)
Do you have any idea to fix this?
Thanks :)

steven-tey:
@daun-io great catch, just fixed this in the PR! Basically you need to make sure that `timeseries_node` is the node that is published in the Tinybird API:

CleanShot 2023-12-11 at 15 44 08

Issue #479, opened 2023-11-28
[ feature ] add option to noindex short links


rishimohan:
I have been using [http://dub.co](https://t.co/5MM1spavfU) for creating short links for my app Pika, the issue i'm facing is that since it's as subdomain it's being indexed by Google and I'm getting warnings in Google Search console. These short links are not in the sitemap or anywhere, but since these links are there on the home page I assume Google must've crawled from there

Is there any way I can prevent my short urls to be indexed? If not it would be great to have a noindex option for custom domain short links on [http://dub.shco(https://t.co/5MM1spavfU)

Yh010:
@rishimohan these two links might help you:

https://chat.openai.com/share/fdf54e64-24a4-44e0-992a-23810fbdfa14

https://knowledge.hubspot.com/seo/prevent-content-from-appearing-in-search-results#:~:text=You%20can%20prevent%20new%20content,to%20the%20content's%20head%20HTML.

Yh010:
@steven-tey  does dub use bitly under the hood?
I am asking this because I am searching for the URL shortener's main script to add the indexing prevention option.

steven-tey:
@Yh010 no it does not – that is an entirely unrelated route for users to authenticate with Bitly for the [Bitly importer](https://dub.co/help/article/migrating-from-bitly).

To customize this indexing behavior, we'll need to programmatically generate [this robots.ts file](https://github.com/steven-tey/dub/blob/main/apps/web/app/robots.ts).

Yh010:
@steven-tey does this look good?
https://github.com/steven-tey/dub/pull/493

steven-tey:
This should be fixed now! :)

rishimohan:
@steven-tey would the links now be noindex by default or is there any option for it?

steven-tey:
they will all be noindex by default (except for the root domain) – we haven't had users ask for option to index short links, but if we do get more of it in the future we'll add an option for it

PR #493, opened 2023-12-11
Update robots.ts for feature request #479

Yh010:
this feature allows for user defined options for indexing/no-indexing. Modified the robots.ts file here.

for taking the user input,we can add a form component in the frontend,so as to define the allowedRules, disallowRules,sitemap in the proposed code


steven-tey:
Thanks for this, but we'll be fixing this in another PR! 🙏

Issue #431, opened 2023-10-26
Hyperlink Links in Comments


abass:
It would be super helpful if links within comments were clickable—ideally not even requiring entering https:// but open to having that if it makes this easier 🔗
![CleanShot 2023-10-26 at 15 43 14](https://github.com/steven-tey/dub/assets/5505994/bc87744f-be88-45f5-b315-c2237bf59fb0)

Sahil-B07:
how to reproduce it ?

aakash2330:
can you assign this to me ?

abhishekgusain07:
is this issue still open ?

steven-tey:
Sorry for the late response, but just assigned this to you @aakash2330! 🙏

Ankitchaudhary810:
i want to ask is that where i can find the problem can you give me a brief @abass like in which section of site ?? 
![image](https://github.com/dubinc/dub/assets/124044913/5f213dbd-db01-486e-8449-cceb011367c2)

steven-tey:
@Ankitchaudhary810 this can be found under the link comments feature on Dub: https://dub.co/help/article/how-to-create-link#link-comments

Ankitchaudhary810:
@steven-tey i understood can you assign it to me?

aakash2330:
@steven-tey raised the required pr https://github.com/dubinc/dub/pull/504

Ankitchaudhary810:
i am facing a problem while setup the project locally is there any separate guide to set up project locally not self hosting guide.

aakash2330:
hey @Ankitchaudhary810 just follow to same steps in the self hosting guide to set it up for local development too .

PR #504, opened 2023-12-17
Linkify comments

aakash2330:
Hyperlink Links in Comments #431

Turn Links within comments clickable . 
(Please ignore the first commit )


apps/web/ui/links/linkify-comments.tsx
```diff
@@ -0,0 +1,23 @@
+
+//to convert links in comments clickable
+export default function LinkifyText({ children }:{children:string}) {
+  const urlPattern = /(\bhttps?:\/\/[-A-Z0-9+&@#/%?=~_|!:,.;]*[-A-Z0-9+&@#/%=~_|])/gi;
+
+  const renderText = (text:string) => {
+    const parts = text.split(urlPattern);
+    return parts.map((part, index) => {
+      if (part.match(urlPattern)) {
+        return <a href={part} key={index} target="_blank" rel="noopener noreferrer" className="hover:underline">{part}</a>;
```
tuffz: I would also recommend adding a "nofollow" to the `rel` attribute to avoid link spammers.

unrenamed (changes requested):
Hey @aakash2330, excited to help with your PR code review!

1) Thinking about making the `<LinkifyText />` component even more modular by moving it to the `packages/ui` module. What do you think? Also, we can tweak the filename to match the component name, maybe something like `linkify-text`.

2) Noticed the #431 issue author mentioned an ideal scenario: allowing URLs without protocols to transform into hyperlinks. 

   > ideally not even requiring entering https:// but open to having that if it makes this easier

    Currently, it seems to support only `http(s)` protocols. Any challenges with this? Maybe check out existing solutions like [linkify.js.org](https://linkify.js.org/)? Let me know your thoughts! 😊

aakash2330:
Hey @tuffz @unrenamed thanks for reviewing my pr . 

1 . Added the "nofollow" to the rel attribute as suggested by @tuffz .
2 . Tweaked the filename to match the component name (linkify-comments) and moved it to packages/ui module .
3. Want to get up a heads up from the maintainers before including a third party library that's why I didn't include it in the first place .

steven-tey:
Thanks a lot for the PR @aakash2330 + feedback from @tuffz & @unrenamed 🙏 

re: `linkifyjs` – it looks lightweight enough, let's use it 😄 (and rely on @socketdev for 3rd party deps security 🔒)

aakash2330:
@steven-tey added linkifyjs to support more URL protocols as well as allowing URLs without protocols to transform into hyperlinks.

steven-tey:
Great work on this PR @aakash2330 – also shoutout to @unrenamed for the linkify tip as well!

Will be merging this now – also updated our [changelog](https://dub.co/changelog/link-comments) & [help docs](https://dub.co/help/article/how-to-create-link#link-comments) to include this too :)

Issue #506, opened 2023-12-17
[FEAT]: `dub` CLI


steven-tey:
Building on @sujjeee's incredible work: https://github.com/sujjeee/dubco

## Proposal 

`npm i dub-cli`

Add an alias from `dub-cli` to `dub`

## Commands

- `dub help`: Shows a list of commands that are available
- `dub login`: [Login to Dub.co](https://github.com/sujjeee/dubco#initialize-config) via API token. 
	- Bonus: Auto-open a new window and log user in, but would require programmatic API token generation. Example: https://github.com/unkeyed/examples/tree/main/unkey-cli
- `dub config`: [Show current config](https://github.com/sujjeee/dubco#show-config)
	- Current user email
	- Current project
	- Current domain
-  `dub config set`: Set current config
	- Project selector
	- Domain selector
	
	Example:

	```
		? Switch to:
	  ── Projects ────────────────────────
	○ Dub (dub)
	○ Steven Tey (steven)
	○ One Word Domains (owd)
	  ─────────────────────────────────
	○ Cancel
	```
- `dub shorten [url] [key]`: [Shorten a URL](https://github.com/sujjeee/dubco#create-shortlink)
	-  If `url` and `key` are not provided, go through the current flow:

		```
		✔ Enter your Destination URL : … 
		✔ Enter your Short link : …
		✔ Short link created!
		```
	- If `url` is provided, create short link with random key
	- If both `url` and `key` are provided, create short link with provided key

## MVP notice 

This is the early MVP of the `dub` CLI and will be improved in the future!

sujjeee:
🙋‍♂️

VishalMCF:
@steven-tey @sujjeee I have one idea related to this. A few days back there was an issue created by you @steven-tey  to programmatically create the QR Code which I wanted to work on. Do you think it will be a good idea to let the user create the QR code by adding some extra parameters and let it download in the user file path? 
From there he can also then do many things like storing it in his S3 or Redis cache. 
What do you guys think?

sujjeee:
https://github.com/dubinc/dub/assets/101963203/5c033536-f6c3-424b-94f0-88e6f01d42e4

sujjeee:
@VishalMCF 👀


https://github.com/dubinc/dub/assets/101963203/594a26c2-a53d-4968-a543-e90a144e106c

VishalMCF:
@sujjeee  Great!! You stole my potential issue already :)

sujjeee:
[...]

PR #510, opened 2023-12-20
Added Command Line Interface (CLI)

sujjeee:
As @steven-tey wrote in a proposal https://github.com/dubinc/dub/issues/506#issue-2045388671 about dub cli, here it is.

### Quick demo

https://github.com/dubinc/dub/assets/101963203/9c941f52-0333-4f38-8411-784b0af33dfd

### Requirement
To publish the package to npm registry, you need an npm `access token`. Create a granular npm access token and add it to your GitHub secret with the name `DUB_AUTH_TOKEN`.

### DIFF
--- a/apps/web/app/app.dub.co/(dashboard)/[slug]/settings/referrals/stats.tsx
+++ b/apps/web/app/app.dub.co/(dashboard)/[slug]/settings/referrals/stats.tsx
@@ -90,7 +90,7 @@ async function StatsInner({ slug }: { slug: string }) {
           label="Affiliate Earnings"
           graphic={<MiniAreaChart data={sales} />}
         >
-          <CountingNumbers prefix="$" fullNumber>
+          <CountingNumbers prefix="$" variant="full">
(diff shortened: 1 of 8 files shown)


sujjeee:
### [ CLI feat. ] : handle prompt onCancel 

https://github.com/dubinc/dub/assets/101963203/8a575a73-4754-4188-9006-ede8c0f6d818

sujjeee:
### [CLI feat.] : added create project command
Previously, when we tried to log in using dub cli, it would show an error if the user did not have any project.

Screenshot 2023-12-23 113916

I fixed this error so that now the user can log in without any project requirement. However, this is only for the login process. If the user does not have any project and logs in, the config file will set the project slug and domain slug/name to null.

When the user creates a new project with the `dub create project` command, it will automatically update the config file with the project slug and domain slug/name based on the project information entered by the user.

### Quick Demo
https://github.com/dubinc/dub/assets/101963203/959ef9b5-112a-4943-84b0-859d0495e172

PR #516, opened 2023-12-23
Added custom props for Icons to change width , height or fill Color o…

ammarrqureshi:
…n the fly.

added CopyIconProps type for accepting props like width, height, and fill Color for the icon in case needed on the fly. All props are optional with default value set as previous.


steven-tey:
Closing this for now since it's not really needed at the moment – but thanks for the PR! 🙏

ammarrqureshi:
Thanks anyway. Can you suggest any thing where I can contribute at the moment? I can do any frontend related contribution

PR #524, opened 2023-12-28
Fix: Dub branding on password protected page

jackblatch:
Issue: #525

Removes `Dub.co` from being appended if the project is on a paid plan on the password protected page.


steven-tey:
Great catch, thank you!

PR #557, opened 2024-01-17
Fix docs

karan0handa:
### Description
`guide.mdx` incorrectly states that `QSTASH_URL` needs to be copied. The `QSTASH_TOKEN` is needed, not `QSTASH_URL`.


steven-tey (approved):

steven-tey:
Good catch, thanks!

PR #558, opened 2024-01-18
fix: fix migration notice on mobile

Snazzah:
Fixed a small issue for the migration notice to show improperly on mobile devices.

| Before | After |
|-|-|
|  |  |


steven-tey:
Amazing, thank you!

PR #570, opened 2024-01-19
Make clicks bar chart adapt to the screen size

### LINKED ISSUE #564: UI bug in Analytics Page in narrow screens
Due to the barchart taking up more width than the screen width, the layout is being broken and allowing to scroll horizontally as if no viewport provided.
image
image

To reproduce, use any device with width less than 350px, then directly go to analytics page. 

Gets fixed when resized gradually width in Chrome Inspect tools.
probably the issue is in useMediaQuery hook while calculating the initial width.

unrenamed:
This fix resolves #564 

TODO:

- [x] upgrade @visx packages versions, also remove unused packages;
- [x] make <BarChart /> a reusable component for rendering any timeseries data, not "clicks" only;
- [x] show "No data available" message when Tinybird returns empty data set;
- [x] add support for mobile view by reducing labels font sizes, changing angles, etc.;
- [x] refactor tooltip to take less space by removing irrelevant margins/paddings;


unrenamed:
### Problem
The bar chart component extends beyond the viewport width **on smaller screens**, causing the page to acquire a horizontal scroll and lose responsiveness. Refer to the screenshots provided by the OP in issue #564 for visual context.

### Root cause (step by step)
1. Say, we open the analytics page on a device with `375px` screen width.

2. On the initial render, the `useMediaQuery` returns `undefined` screen width.
https://github.com/dubinc/dub/blob/c74c8b54bd607fff78243d3843e6035eab2bf830/apps/web/ui/stats/bar-chart.tsx#L41

3. Since, `screenWidth` is `undefined` now, `CHART_WIDTH` gets calculated using the `CHART_MAX_WIDTH` value which is `800px`.
https://github.com/dubinc/dub/blob/c74c8b54bd607fff78243d3843e6035eab2bf830/apps/web/ui/stats/bar-chart.tsx#L43-L51

4. `CHART_WIDTH` is currently set at `800px` to determine the width of the chart, inner grid, and other elements, effectively expanding the layout viewport.

However, in subsequent re-renders, the `useMediaQuery` returns `842px` instead of the expected `375px` or less. While this may seem unusual, there is no bug in the hook implementation; it accurately reflects the value of `window.innerWidth`.
  
Let's check the [definition of the innerWidth property](https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth):
> The read-only [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window) property innerWidth returns the interior width of the window in pixels (that is, the width of the window's [**layout viewport**](https://developer.mozilla.org/en-US/docs/Glossary/Layout_viewport))

The key words here are **layout viewport**. From MDN:
> The layout viewport is the viewport into which the browser draws a web page. Essentially, it represents what is available to be seen, while the [visual viewport](https://developer.mozilla.org/en-US/docs/Glossary/Visual_Viewport) represents what is currently visible on the user's display device.

tl;dr; `innerWidth/innerHeight` reflect viewport with the pinch-zoom factor applied and the page is loaded fully zoomed out. This is why we get `842px`.

5. Now, by applying an 80% factor to the `842px` value (as outlined in step 3), we arrive at a new `CHART_WIDTH` approximately equal to `670px`. This remains nearly twice the width of the screen.
image

### Fix
The fix is to use a more idiomatic way to handle screen size changes and make the chart to adapt to it — [@visx/responsive](https://airbnb.io/visx/docs/responsive). This package offers an API for creating responsive graphs, bringing several key advantages:
1. Chart width is no longer tied to arbitrary size factors or constants (such as MAX_WIDTH);
2. Significantly reduces overall component complexity by delegating resize logic to a dedicated component;

unrenamed:
@steven-tey 👋 This PR is ready for code review

devkiran:
Thanks for your PR. We shipped a new version of the analytics UI https://d.to/analytics

Issue #567, opened 2024-01-18
Fix all UI Components from @dub/ui being client components


versecafe:
## Source of Problem
```ts
import { defineConfig, Options } from "tsup";

export default defineConfig((options: Options) => ({
  entry: ["src/**/*.tsx"],
  format: ["esm"],
  esbuildOptions(options) {
    options.banner = {
      js: '"use client"',
    };
  },
  dts: true,
  minify: true,
  external: ["react"],
  ...options,
}));
```

Currently this is needed to prevent Next from crashing whenever a component that would normally need the "use client" directive is called from the package but as tsup 6.1.3 does not support file level directives it has to be injected onto all the files, this is currently increasing first load JS across the project.

To fix this there needs to be a way of bundling client and server components separately, currently following this issue https://github.com/egoist/tsup/issues/835 as it proposes some solutions such as splitting pure RSCs and then mixed + client components into separate bundles, second option which may be more effective but won't produce exportable packages for npm is to have Next transpile the packages https://nextjs.org/docs/app/api-reference/next-config-js/transpilePackages.

steven-tey:
Really great feedback here, thank you @versecafe! Excited to see that PR develop 👀

versecafe:
@steven-tey This does make create some annoying behaviour for an external user using the package on something other than NextJS but as next is actively in the deps list that should be extremely unlikely

## New addition to package readme
### Transpiling

As of version `0.1.0` the package is not bundled, this means transpilining the module is 
highly recommended, this was done to support React Server Components.

If you are using NextJS this is as simple as:

```js
// next.config.js
/** @type {import('next').NextConfig} */
module.exports = {
  transpilePackages: ["@dub/ui"]
}
```

versecafe:
If there are any changes you'd like let me know but it should be production ready, and got a version bump as it would be considered a breaking change for someone using it externally. It is compatible and fully functional with the current dub/main branch and there should be no need for any change for any branches besides if someone has ui changes in work they'll need to ensure they properly declare client components.

PR #572, opened 2024-01-19
[FEAT]: use RSCs in @dub/ui Fixes

versecafe:
# Fixes https://github.com/dubinc/dub/issues/567

# Goal
Reduce client side load and improve initial input delay times by taking advantage of RSCs in Next rather than having tsup apply a `"use client";` banner over the entire package. Relies on 

```js
/** @type {import('next').NextConfig} */
module.exports = {
  transpilePackages: ["@dub/ui"],
  ...
}
```

> This can be continually improved by more carefully crafting components to minimize client code inside in things such as the Nav & Suspense boundaries setup in preparation for ppr becoming stable.


versecafe:
@steven-tey This should be good to go, everything compiles and works in local testing without issue and all components that are portable to RSCs without a major refactor have been moved.

versecafe:
@subh-cs When you have some time could you review this and what's your opinion on bumping the package version to 0.1.0 as the transpile step is a breaking change for any external user *not that I think there are any but just in case*

steven-tey:
This looks great, thank you! Curious to know what the approximate improvement in initial input delay time would be with this change – were you able to run any benchmarks for this?

versecafe:
@steven-tey Not sure on the exact reduction testing FID has been a bit annoying, but when using artificial throttling for "low end mobile" on chrome dev tools there was a decent improvement in page load and responsiveness, This should be stable to merge but also could use improvements on splitting bigger portions like the Nav & Footer into client and server components before hand if that would be preferred.

versecafe:
### Reactions to setting off a giant socket security report without ever adding a new package

1. break lock file during tiny merge conflict
2. "this is fine just run `pnpm i`
3. "oh no oh no this just flagged dozens of package changes"
4. realize it's just minor version bumps allowed by the package json specifications

versecafe:
@steven-tey would this be good to merge as all components are now client or server and properly being bundled as RSCs and client components or do you want me to go break down some of the larger portions like Nav and Footer into client server splits since they are larger components that load on most every page?

PR #585, opened 2024-01-26
REST API

devkiran:
This PR add API server into a standalone app (Hono). Still working in progress.


README.md
steven-tey, line 1: What is changed here? 👀

apps/api/lib/projects.ts
```diff
@@ -0,0 +1,146 @@
...
+  //       plan: project.plan,
+  //       limit: project.linksLimit,
+  //       type: "links",
+  //     }),
+  //     {
+  //       status: 403,
+  //       headers,
+  //     },
+  //   );
+  // }
```
steven-tey, line 133: Why are these commented out?

PR #587, opened 2024-01-29
Add QR Code API

### LINKED ISSUE #452: Generate QR code programmatically with API
Would love the ability to generate & download QR codes programmatically via the API.

steven-tey:
Closes #452


versecafe:
Just a warning on the way rate limiting is being done, it itself is vulnerable to a DDoS as if someone hammers your API enough to push past the daily command limit on free tier or your commands per second *likely 1000 which should not be difficult in a small bot net* on any tier it will cause all rate limited APIs to throw errors and fail to function.

## Example
image

Not the Dub API but discovered this using a similar upstash rate limiter on my own Next API on Vercel and found the cause.

PR #592, opened 2024-01-30
API Changes

devkiran:


versecafe:
@devkiran Would moving to drizzle make sense if already adding zod wrappers? fix prismas performance issues and [dirzzle-zod](https://orm.drizzle.team/docs/zod) makes fusing the two to generate type validation from db models easier

apps/docs/api-reference/errors.mdx
steven-tey, line 1: Can we add the status codes for each of these responses here? E.g. 403, 404, 400, etc.

apps/web/app/api/analytics/[endpoint]/route.ts
Comment (steven-tey) [inline apps/web/app/api/analytics/[endpoint]/route.ts:1]: this looks great – can we get this added to `/api/analytics/edge/[endpoint]/route.ts` as well?

apps/web/app/api/cron/links/event/route.ts
steven-tey, line 1: Why do we need to include `excludeRoot` here if it's an optional param?

apps/web/app/api/links/route.ts
```diff
@@ -1,89 +1,106 @@
...
+          throw new DubApiError({
+            code: "rate_limit_exceeded",
+            message:
+              "Rate limited – you can only create up to 10 links per day without an account.",
+          });
+        }
       }
-    }
 
-    const { link, error, status } = await processLink({
```
steven-tey, line 59: what happened to `status`? should we remove it and fully rely on `code`?

apps/web/app/api/og/analytics/route.tsx
```diff
@@ -36,6 +36,7 @@ export async function GET(req: NextRequest) {
     linkId: link.id,
     endpoint: "timeseries",
     interval: "30d",
+    excludeRoot: false
```
steven-tey: same question about `excludeRoot` – why do we need to define it? can it just be left undefined

steven-tey (changes requested):

apps/docs/api-reference/errors.mdx
devkiran: Yup, I'll add status codes.

apps/web/app/api/analytics/[endpoint]/route.ts
Comment (devkiran) [reply apps/web/app/api/analytics/[endpoint]/route.ts:1]: Sure.

apps/web/app/api/cron/links/event/route.ts
devkiran: Good catch, can fix it with some tweak in zod schema.

apps/web/app/api/og/analytics/route.tsx
devkiran: Yup, it can left undefined

apps/web/app/api/links/route.ts
devkiran: `status` was replaced with error `code`, no longer needed. I think we can safely remove `status`.

steven-tey (approved):

PR #621, opened 2024-02-21
Metatags API rate limit increase

jparkrr:
Hi there!

Can you increase the rate-limit of the metatags api to match what's listed on https://dub.co/docs/api-reference/rate-limits?

I'm not hitting it continuously for my use-case, but when a user loads a page with a lot of links to be processed I'd like to be able to show the previews somewhat quickly.

Thanks!


steven-tey:
Not comfortable bumping this up since this is a free service and I'm worried about folks abusing it 😅 

The code is fairly simple though so you can easily extract it into a simple Next.js App on Vercel and it should work fine :)

PR #624, opened 2024-02-22
feat: add lock icon to protected link cards

Snazzah:
Small change, helps differentiate links more similar to cloaked links.
image


steven-tey:
Great idea, thank you!

PR #636, opened 2024-02-28
google and github login button loading fixed

nitishmahawar:
After clicking google or github login button in login and register page it redirects to auth page but when we click back  button is still in loading state and other buttons in disabled state. I fixed this issue.
 
Screenshot (172)


steven-tey:
Good catch, but we're gonna rely on `useEffect` instead to prevent an interim non-loading state

nitishmahawar:
> Good catch, but we're gonna rely on `useEffect` instead to prevent an interim non-loading state

Sorry but `useEffect ` is not gonna work in this case

PR #637, opened 2024-02-28
Affiliate tracking

zlwaterfield:
- Added new Tinybird data source for conversions
- Added tracking endpoints for clicks and conversions
	- Two API endpoints added `/api/track/click` and `/api/track/conversion`
	- Both are using API key authentication with a new withAuthEdge handler - note: need to test this + the rate-limiting logic
	- Both are using Zod for validation
	- The click endpoint takes in a URL, finds the link, records a click on that link with that Click ID and returns a Click ID to the user
	- The conversion endpoint takes in the event name, properties and click id and records the conversion in the new Tinybird data source
	- The Click ID will be recorded across the clicks and conversions so we can track the lifecycle of a user events
- Added affiliates tables to Prisma.
	- A company will create affiliates in the dashboard (UI not done in this PR)
	- Affiliates will belong to a Project (and Campaign - though they aren't live yet) with an optional User
	- When a campaign is setup it will have a `domain`, `url` and `affiliate param key` associated, when the affiliate is created it will have a `username`. So each affiliate will have a URL generated either as a short link like `https://${campaign.domain}/${affiliate.username}` or as the full URL like `https://${campaign.url}?{campaign.affiliate_param_key}=${affiliate.username}`. Note: the campaign URL is really just the root domain of their website. This is a workaround when they don't want to use a short link. And if they don't want to use a short link, the domain will equal the URL.
		- When `https://${campaign.url}?{campaign.affiliate_param_key}=${affiliate.username}`  is used the SDK will auto track the click since the `campaign.affiliate_param_key` is in the URL.
		- When `https://${campaign.domain}/${affiliate.username}` is used it will redirect to `https://${campaign.url}?{campaign.affiliate_param_key}=${affiliate.username}`. A click will be tracked on the redirect as well as the ending url visit with the SDK, but it will be deduced on the server.  In this case, the SDK will not overwrite the click ID in the cookies.
- Updated the analytics SDK to support all of the above.

Note: this is focused on the affiliate model. This doesn't include support for non-affiliate tracking for simple conversion tracking on all links. 

This isn't foolproof, so call me out if I'm missing anything.

❓Questions ❓
- For the conversion events API endpoint, how do we get the affiliate? We don't have a link or a project, and the API key is generic to the user. 
- Do we need to be passing the affiliate username in the click and conversion events? Is there another way to get it? Right now we pass affiliate param key on the click events and affiliate username on the conversion events. 

🗒️ Things to do 🗒️
- Add the Campaign model
- Add the UI portion so companies can create/configure campaigns, create affiliates and view the analytics for campaigns and affiliates
- Figure out a way to associate Affiliates with Users


apps/web/app/api/track/click/route.ts
```diff
@@ -0,0 +1,54 @@
...
+  } else {
+    link = await redis.hget<RedisLinkProps>(domain, key);
+    affiliateUsername = key
+  }
+  if (!link) {
+    return NextResponse.json("Link not found", { status: 404 });
+  }
+  const clickId = generateClickId();
+  const affiliate = await getAffiliateViaEdge(link.project_id, affiliateUsername);
+  const affiliateId = affiliate?.id;
```
steven-tey: Since we're only returning the clickId, let's move all of these await calls into the `waitUntil` so it doesn't block the response? Ideally we'd want the response to be instantaneous

apps/web/app/api/track/conversion/route.ts
```diff
@@ -0,0 +1,39 @@
...
+// POST /api/track/conversion – post conversion event
+export const POST = async (req: NextRequest) => {
+  const body = ConversionEventSchema.parse(await req.json());
+  const { eventName, properties, clickId, affiliateUsername } = body;
+
+  let affiliateId;
+  if (affiliateUsername) {
+    const affiliate = await getAffiliateViaEdge('link.project_id', affiliateUsername);
+    affiliateId = affiliate?.id;
+  }
```
steven-tey: Same here, let's move all of these await calls into the `waitUntil` so it doesn't block the response

apps/web/lib/auth-edge/index.ts
```diff
@@ -0,0 +1,99 @@
...
+            hashedKey,
+          },
+        },
+      },
+      select: {
+        id: true,
+        name: true,
+        email: true,
+      },
+    });
```
steven-tey: Don't think we can use Prisma functions here since it's edge runtime

apps/web/lib/auth-edge/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      });
+    }
+    await prisma.token.update({
+      where: {
+        hashedKey,
+      },
+      data: {
+        lastUsed: new Date(),
+      },
+    });
```
steven-tey: Ditto the above:

> Don't think we can use Prisma functions here since it's edge runtime

apps/web/lib/planetscale.ts
```diff
@@ -61,6 +61,54 @@ export const getLinkViaEdge = async (domain: string, key: string) => {
...
+        password: string | null;
+        expiresAt: string | null;
+        ios: string | null;
+        android: string | null;
+        geo: object | null;
+        projectId: string;
+        publicStats: number;
+      })
+    : null;
+}
```
steven-tey: FWIW we're planning to move this to Redis in the future to avoid overloading our MySQL DB. We'll probably create separate KV pairs (on top of our existing ones that's indexed on the domain+key) for this.

steven-tey (changes requested):

apps/web/app/api/track/click/route.ts
zlwaterfield: Updated!

apps/web/app/api/track/conversion/route.ts
zlwaterfield: Updated!

apps/web/lib/auth-edge/index.ts
zlwaterfield: Updated!

apps/web/lib/auth-edge/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      "1 s",
+    ).limit(apiKey);
+
+    const headers = {
+      "Retry-After": reset.toString(),
+      "X-RateLimit-Limit": limit.toString(),
+      "X-RateLimit-Remaining": remaining.toString(),
+      "X-RateLimit-Reset": reset.toString(),
+    };
+    console.log('success', success, 'limit', limit, 'reset', reset, 'remaining', remaining)
```
adasarpan404: we can remove this logs. it is good for debugging.

apps/web/lib/auth-edge/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    req: NextRequest;
+    params: Record<string, string>;
+    searchParams: Record<string, string>;
+    headers?: Record<string, string>;
+    session: Session;
+  }): Promise<Response>;
+}
+
+export const withAuthEdge =
+  (handler: WithAuthHandler, {}) =>
```
adasarpan404: The empty object {} in the `withAuthEdge` function. What is the purpose of that?

apps/web/lib/auth-edge/index.ts
zlwaterfield: Good catch!

apps/web/lib/auth-edge/index.ts
zlwaterfield: It was to copy the structure of `withAuth` but agree it's not needed. Will remove.

PR #649, opened 2024-03-04
Add warning if apiKey is provided but no projectSlug

steven-tey:


zlwaterfield (approved):
Nice work, this is a great improvement!

PR #674, opened 2024-03-11
Docs/GitHub app instructions

baristaGeek:
I was having some trouble with the GitHub application setup step for developing locally. 

I added the steps that I used to get this working as instructions. Hope it makes it more friendly for more contributors.


steven-tey:
Thank you for this! We moved our docs repo to https://github.com/dubinc/docs – do you mind making the same PR there instead?

PR #688, opened 2024-03-15
Move from Cloudinary to a storage bucket solution

steven-tey:
This PR is migrating from Cloudinary to a storage bucket solution. The integration is for S3 but we will be using R2 which is an S3 compatible solution.


apps/docs/images/cloudflare-r2-manage-api-tokens.png
zlwaterfield, line 1: @steven-tey how did you edit this image?

zlwaterfield, line 1: @steven-tey how did you edit this image?

apps/web/lib/api/projects.ts
```diff
@@ -128,11 +126,9 @@ export async function deleteProjectAdmin(
   ]);
 
   const deleteProjectResponse = await Promise.all([
-    // delete project logo from Cloudinary
-    project.logo &&
-      cloudinary.v2.uploader.destroy(`logos/${project.id}`, {
-        invalidate: true,
-      }),
+    // delete project logo if it's a custom logo stored in R2
+    project.logo?.startsWith(`https://${process.env.STORAGE_DOMAIN}`) &&
```
zlwaterfield: I think we should be consistent with `STORAGE_DOMAIN` and `STORAGE_ENDPOINT` in terms of whether or not it has the `https` protocol on it. `STORAGE_ENDPOINT` will always have it so we should make `STORAGE_DOMAIN` also have it. Since domain isn't the right now I can rename to `STORAGE_BASE_URL`

apps/docs/images/cloudflare-r2-manage-api-tokens.png
steven-tey: I used [Cleanshot](https://cleanshot.com/)! :)

apps/web/lib/api/projects.ts
steven-tey: agreed, I think having `STORAGE_BASE_URL` makes much more sense here :thumbs-up:

PR #704, opened 2024-03-23
Update toggle.tsx

destroyer743:


steven-tey:
we updated our domain filter recently so this is no longer needed, but thank you!

PR #708, opened 2024-03-24
Fix link creation UI hanging in Safari

### LINKED ISSUE #684: Creating a new link results in the UI hanging on the "loading" state
### Repro steps

1. At `https://app.dub.co/michael-flores` I click the button _Create link_
2. In the form, I paste in some destination
3. I click the button at the bottom of the form _Create link_
4. I observe the UI hangs at this stage, no matter how long I wait
5. If I refresh, the link was created successfully. It does this for every link I create. 

### Resources
In the console, step 3 results in the output:
```
Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
```

Browser: Safari Version 17.4

https://github.com/dubinc/dub/assets/11711706/5578dabc-5fbc-412a-bca0-2a69d1957aa3

### Workaround available

1. When creating link, user can refresh pretty much right away (or however long it takes their network to send the POST), and see the link instead of waiting.

TWilson023:
Safari seems to have an issue with using `navigator.clipboard` to copy text without it being the direct result of a user action. In this case waiting for a request before copying the shortlink seems to be the issue.

We could probably investigate some other solutions, maybe even optimistically copying, but for now this seems to fix the UI hanging in Safari.

Closes #684

### DIFF
--- a/apps/web/ui/modals/add-edit-link-modal/index.tsx
+++ b/apps/web/ui/modals/add-edit-link-modal/index.tsx
@@ -351,8 +351,8 @@ function AddEditLinkModal({
                     router.push("/links");
                     setShowAddEditLinkModal(false);
                   }
-                  // copy shortlink to clipboard when adding a new link (if document is focused)
-                  if (!props && document.hasFocus()) {
+                  // copy shortlink to clipboard when adding a new link
+                  if (!props) {
                     try {
                       await navigator.clipboard.writeText(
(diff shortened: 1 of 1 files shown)


steven-tey (approved):
Great fix, thank you! Would definitely love to explore optimistically copying next! 😄

PR #710, opened 2024-03-27
ENG-136: Add time zone handling to time series analytics

TWilson023:
Adds time zone parameter to `/api/analytics` endpoint and passes it to TinyBird for use in `toStartOf...` functions so that the granularity periods line up with the requestor's local time.


apps/web/ui/analytics/clicks-chart.tsx
```diff
@@ -62,7 +62,8 @@ export default function ClicksChart() {
                 <strong className="text-gray-800">
                   {nFormatter(d.values.clicks, { full: true })}
                 </strong>{" "}
-                clicks
+                click
+                {d.values.clicks !== 1 && "s"}
```
TWilson023, line 66: Sneaking in this little fix as well!

steven-tey: Amazing!

steven-tey:
will be added in #915

PR #718, opened 2024-03-30
feat: add posthog

zlwaterfield:
Adding Posthog. 

This first PR is adding page views and autocapture and identify. In subsequent PRs we will add tracking for custom events.

Note: this is not using a proxy right now. We will need to add that.

related to https://github.com/dubinc/dub.co/pull/30


apps/web/app/providers.tsx
```diff
@@ -0,0 +1,15 @@
+'use client'
+
+import posthog from 'posthog-js'
+import { PostHogProvider as PHProvider } from 'posthog-js/react'
+
+if (typeof window !== 'undefined') {
+  posthog.init(process.env.NEXT_PUBLIC_POSTHOG_KEY as string, {
+    // api_host: process.env.NEXT_PUBLIC_POSTHOG_HOST as string,
```
zlwaterfield, line 8: I'm finding Next.js proxies to be flakey when using rewrites, and I can't seem to get the middleware to hit. Since we don't have the domain in Cloudflare, we can't use a CF worker proxy. @steven-tey maybe you can help me get the middleware to hit and we can follow these https://posthog.com/docs/advanced/proxy/nextjs-middleware

Issue #701, opened 2024-03-21
Add support for '_' to allow migration from other URL Shorteners


miles202:
Dub does not allow for underscores ('_') in the shortened name, which prevents migration from other shorteners that do support it.

steven-tey:
@miles202 thanks for the feature request! Do you know which URL shorteners support `_`? AFAIK the main ones (Bitly, Rebrandly, etc.) don't support it

miles202:
We are migrating from www.rebrandly.com and have one...just added a new one to test:

https://fieldhub.link/test_one

It allows it and it works.

steven-tey:
@miles202 we could potentially allow `_`, but not as the first character in the key (e.g. `fieldhub.link/_test` wouldn't work). Would that work for you?

miles202:
Yes that would work perfectly but I am personally ok without it. The few cases where we have it I can change to dash and re distribute the url since they are mostly internal. I just submitted the ticket to make sure there was good support for migrations from the legacy platforms generally. Thanks!!

vagishfibr:
@steven-tey can u please assign the issue to me, code change identified

steven-tey:
Underscores is now supported on Dub links 🚀 

![CleanShot 2024-08-19 at 13 03 54](https://github.com/user-attachments/assets/91132f2c-aa50-494f-9f5b-d11f2e8e68b3)

PR #720, opened 2024-03-31
Modify regex on UI and Utils to Allow Underscores

dilawari2008:
Solves #[701](https://github.com/dubinc/dub/issues/701)


steven-tey:
Added support for this via #1202

PR #721, opened 2024-04-01
Introduce Empty Filter State

flornkm:
This PR introduces an empty state for properly routing the user to the correct modal for creating filter options.

### Demo
https://github.com/dubinc/dub/assets/92092993/1d55cd28-f108-494b-8347-88b39a77354b


steven-tey (approved):

steven-tey:
Awesome, thank you!

PR #728, opened 2024-04-02
Update link info api response

devkiran:
- Add missing attributes (`workspaceId`, `tags`, `qrCode`, etc) to the link info response.
- Move the link response transformation logic to the utils.


steven-tey (approved):
Great work, love the new `transformLink` helper function 🙌

PR #739, opened 2024-04-06
Transfer domain to another workspace

devkiran:


apps/web/app/app.dub.co/(dashboard)/[slug]/domains/page-client.tsx
Comment (steven-tey) [inline apps/web/app/app.dub.co/(dashboard)/[slug]/domains/page-client.tsx:1]: Instead of prop drilling here, I believe we can just do `const { activeWorkspaceDomains } = useDomains()` inside the `<DomainCard />` component

steven-tey (changes requested):
aside from the `activeWorkspaceDomains` prop-drilling, LGTM!

steven-tey (approved):

steven-tey (approved):

PR #762, opened 2024-04-11
Fix : Valid Input Field Update in Workspace Settings & Toast Message Update.

anirbanpaulcom:
### Description:
This implementation addresses the issue where the form does permit submissions with name, slug containing only whitespace characters. Additionally, it includes proper error toasts for notification.


steven-tey:
Great catch, thank you!!

Issue #682, opened 2024-03-13
feat - view or add title/description in the individual short link card


ankitsridhar16:
Hello,

When I create a new short link, the title and description are automatically updated based on the original link (as per the response). However, when we save and view the section where all the generated links are listed by card, it currently doesn't show the title or description. This makes it a bit harder to understand what the link is about. It would be beneficial if we displayed the title or description generated in the individual link card or allowed for the creation of a custom description/title for each link.
<img width="910" alt="Screenshot 2024-03-13 at 9 48 24 AM" src="https://github.com/dubinc/dub/assets/36269838/722a64e2-e04f-4050-a0dc-d0aeb0f04ebc">
<img width="820" alt="Screenshot 2024-03-13 at 9 48 43 AM" src="https://github.com/dubinc/dub/assets/36269838/6ba2c947-32c5-4d7c-b1de-7169d2bc8a34">

robkallday:
Yes this would be helpful. Currently I add the title as a comment and need to hover over the comment icon to see the title. Would be nice to see the title when scrolling through all the links.

mohammedyunus2002:
i would like to work on this issue. Can you specify where do you want tile and description to be rendered

steven-tey:
Good news: We just shipped this today! Thanks a lot for all your feedback 🙏 

https://x.com/steventey/status/1816530501919969706

PR #783, opened 2024-04-15
Feat: Added title and description to the link card

ashimgupta39:
#682 
Hi,

I've enabled displaying both the title and description in the link card. The description now appears in a tooltip, visible only when present. I've also ensured the title is shown at the start of the link which is also visible only when present-

Screenshot 2024-04-16 at 12 40 14 AM


steven-tey:
This has been shipped via https://github.com/dubinc/dub/pull/1021

PR #789, opened 2024-04-17
Added start and end date optional params added to analytics endpoint

### LINKED ISSUE #749: Add Start and End date as an optional parameter for all analytics endpoints
Currently it is not possible to aggregate analytics data between two dates in the past.
The [getAnalytics](https://github.com/dubinc/dub/blob/63ca3072928a1ee08dc398d640019ac11e79ed51/apps/web/lib/analytics.ts#L95) query is already using a start and end date so adding this functionality to the API should just be a matter of making the start and end dates dynamic based on the optional params passed in.

UsmanDev09:
fixes #749 

**Description:**

The endpoint `analytics/[endpoint]` now accepts `startDate` and `endDate` as optional params in ISO Date Time Format `yyyy-MM-dd'T'HH:mm:ss. SSSXXX`. I've added few checks too: 

1. Start Date should be after 22 September, 2022
2. End Date cannot be in future
3. Either `interval` or `start and end date` can be provided, not both

P.S. Once the PR is merged, I'll update the docs and add an option in the UI to filter analytics by providing custom dates.


apps/web/app/api/analytics/[endpoint]/route.ts
```diff
@@ -13,9 +13,22 @@ export const GET = withAuth(
...
+    if (
+      (interval && startDate && endDate) ||
+      (startDate && !endDate) ||
+      (!startDate && endDate)
+    ) {
+      throw new DubApiError({
+        code: "unprocessable_entity",
+        message: "Either provide interval or start and end date, not both",
+      });
+    }
```
Comment (anirbanpaulcom) [inline apps/web/app/api/analytics/[endpoint]/route.ts]: Isn't this checking needed? What do you think?
```suggestion
     if (interval || (startDate && endDate)) {
      throw new DubApiError({
        code: "unprocessable_entity",
        message: "Either provide interval or start and end date, not both",
      });
    }

    if (startDate && endDate && new Date(startDate) <= new Date(endDate)) {
      throw new DubApiError({
        code: "unprocessable_entity",
        message: "Start date must be before end date",
      });
    }
```

apps/web/lib/zod/schemas/analytics.ts
```diff
@@ -10,6 +10,36 @@ export const getAnalyticsQuerySchema = z.object({
...
+        const todaysDate = new Date();
+        const inputDate = new Date(value);
+        return inputDate <= todaysDate;
+      },
+      {
+        message: "The end date cannot be in future.",
+      },
+    )
+    .optional()
+    .describe("The end date and time when to retrieve analytics from."),
```
anirbanpaulcom: It will check if the date is between or not. What are your thoughts?

```suggestion
const isBetweenFoundingAndNow = (value: string) => {
  const inputDate = new Date(value);
  const foundingDate = new Date("2022-09-22T00:00:00.000Z");
  const today = new Date();
  return inputDate >= foundingDate && inputDate <= today;
};

export const getAnalyticsQuerySchema = z.object({
  domain: z.string().optional().describe("The domain of the short link."),
  key: z.string().optional().describe("The short link slug."),
  interval: z
    .enum(intervals)
    .optional()
    .describe("The interval to retrieve analytics for."),
  startDate: z
    .string()
    .datetime()
    .refine(isBetweenFoundingAndNow, {
      message: "The start date must be between September 22, 2022 and Now.",
    })
    .optional()
    .describe("The start date and time when to retrieve analytics from."),
  endDate: z
    .string()
    .datetime()
    .refine(isBetweenFoundingAndNow, {
      message: "The end date must be between September 22, 2022 and Now.",
    })
    .optional()
    .describe("The end date and time when to retrieve analytics from."),

```

jmagentio:
@UsmanDev09 this looks great. Is there anything else needed to ship it?

UsmanDev09:
@jmagentio Nope, it's complete

jmagentio:
@steven-tey is this the branch and implementation you are hoping to support start and end dates with for analytics?

steven-tey:
@jmagentio yes we're reviewing this PR currently cc: @devkiran

steven-tey (approved):
Everything LGTM, thanks a lot for the PR!

Will merge tomorrow AM 🙏

UsmanDev09:
Thanks! Do you want me to update the docs and add the option to filter analytics with custom dates in the UI in a separate PR? @steven-tey

anirbanpaulcom (approved):

steven-tey:
@UsmanDev09 docs should be updated automatically! Let's hold off on the custom client side filter for now – we want to explore something that's mobile friendly + supports presets :)

PR #803, opened 2024-04-22
add a way to let users set an `externalId` for a link which allow users to fetch the link based on that.

devkiran:
This will let users pass their own user ID under the `externalId` field when creating a link.

Then, they can:
- Update the link
- Delete the link
- Retrieve analytics for the link 

...simply by passing their external ID (prefixed with `ext_`)


steven-tey (approved):
Fantastic idea! 💡

steven-tey (approved):

PR #807, opened 2024-04-23
Add tests for link redirects

devkiran:


apps/web/tests/redirects/index.test.ts
```diff
@@ -0,0 +1,138 @@
+import { Link } from "@prisma/client";
+import { describe, expect, test } from "vitest";
+import { IntegrationHarness } from "../utils/integration";
+
+const poweredBy = "Dub.co - Link management for modern marketing teams";
+const domain = "dub.sh";
```
steven-tey: this should be the preview URL domain right?

steven-tey (changes requested):
edit: we need to use the right domain

apps/web/tests/redirects/index.test.ts
devkiran: You're right. Updated to use the preview URL.

steven-tey (approved):

Issue #360, opened 2023-09-02
Feature request: Language selection for the 'Password' modal.


patrickmast:
Hello,

Allow me to suggest an enhancement:
Incorporating a language selection feature for the 'Password' modal.
<img width="508" alt="image" src="https://github.com/steven-tey/dub/assets/424387/6928ebe4-ee92-490c-85c2-a9ae21003e9e">

My user base has a limited understanding of English. 😅

Thanks!

steven-tey:
i18n is not very high on the roadmap currently, but it's super important to us so it'll be added very soon!

gcho91:
Side note - the phrasing here may be confusing for non-native English speakers. May I suggest simplying it to:  "This link is password protected" or "Enter the password to view this link"?

sujal12344:
I would like to contribute here. Can you please assign me. @patrickmast @gcho91 ?

PR #826, opened 2024-04-28
Issue#360 Updated description for easier translation

LincolnKne:
For a class I am in we needed to participate on an issue in an open source project, Issue #360 seemed like a temporary fix was possible until I18N is implemented over the whole application. I decided to update the files const description from "This link is password protected. Please enter the password to view it." to "Access to this link is secured with a password. Please enter the password to proceed." as I feel this translates easier to many different languages.


steven-tey:
We'll improve our i18n support soon, but thanks for the effort here!

PR #849, opened 2024-05-02
Validation added for 🗺️

### LINKED ISSUE #780: Some emoji's don't work
Tried creating a link containing the map emoji 🗺️ and got an error that only emojis, -, and / are allowed.

UsmanDev09:
fixes #780 

I have added validation for this emoji since our pattern `[\p{L}\p{N}\p{Pd}\/\p{Emoji}]+
` was failing to recognize it as emoji. However, there might be a lot of similar cases in future since new emojis are regularly added to unicode. 

Using this [Library](https://github.com/mathiasbynens/emoji-regex) is a better solution since it generates the regex pattern at build time based on Unicode. Moreover, it can be updated whenever new emojis are added.


steven-tey:
closing in favor of #1215

PR #852, opened 2024-05-03
Changed max width of screen size so page layout would not side scroll…

Tristan-Sayasit:
Fixing issue #564.


steven-tey:
I believe this has been fixed with our new analytics chart component!

PR #859, opened 2024-05-04
Quick Fix: Inconsistent Error Responses for Invalid JSON in API Requests 

aliamerj:
When sending invalid JSON to **PATCH**  `/links/{linkId}` and **POST** `/links/bulk`, we're getting an _"internal server error"_  instead of the more helpful _"bad request"_ error.
However, **POST** `/links` gets it right and tells us it's a "bad request" due to invalid JSON.

I believe this will help everyone using the API to have a smoother and more predictable experience.

I've already fixed it by first checking:
```ts
const bodyRaw = await parseRequestBody(req);
const links = bulkCreateLinksBodySchema.parse(bodyRaw);
```
I hope this helps make Dub better!

### DIFF
--- a/apps/web/app/api/links/[linkId]/route.ts
+++ b/apps/web/app/api/links/[linkId]/route.ts
@@ -5,6 +5,7 @@ import {
   transformLink,
   updateLink,
 } from "@/lib/api/links";
+import { parseRequestBody } from "@/lib/api/utils";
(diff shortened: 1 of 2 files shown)


steven-tey (approved):

steven-tey:
Brilliant work, thank you @aliamerj!

aliamerj:
Thanks

PR #918, opened 2024-05-22
styling: harmonize colors in date-range-picker to gray-200

christopherkindl:
CleanShot 2024-05-22 at 23 59 26@2x


steven-tey:
Looks like this has been fixed – thank you! 🙏

PR #945, opened 2024-06-01
Show warning when duplicate destination URL exists (#926)

### LINKED ISSUE #926: [FEAT]: Warning popup when duplicate destination URL exists
Via a user request:

> We have a very distributed, remote, volunteer team, and we're trying to more consistently enforce using shortened links so we have better click tracking, but sometimes team members will go in to create a link for something that already has a link, not knowing that it exists. We'd love to have the option to enable a warning popup if the Destination URL specified when creating a new link already exists in another Short Link.

Implementation: Create a new endpoint that checks if a given destination URL exists in a workspace. 

Bonus: Nail down the UI implementation for this – use `AnimatedSizeContainer` for a smooth transition. See Linear's ["similar issues" UX](https://linear.app/changelog/2023-08-03-similar-issues) for reference

kasipavankumar:
Show an error message when destination URL exists for a domain.

## Screenshot

image

Fixes #926

### DIFF
--- a/apps/web/app/api/links/verify/destination-url/route.ts
+++ b/apps/web/app/api/links/verify/destination-url/route.ts
@@ -1,0 +1,32 @@
+import { DubApiError, handleAndReturnErrorResponse } from "@/lib/api/errors";
+import { destinationUrlChecks } from "@/lib/api/links/utils";
+import { domainDestinationUrlSchema } from "@/lib/zod/schemas/links";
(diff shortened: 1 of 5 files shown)


steven-tey:
Thanks for the PR! We've updated our [link builder UI](https://dub.link/builder). Please open a new PR with the latest designs instead 🙏

PR #947, opened 2024-06-02
AI Filters

steven-tey:
✨ 👀 

https://github.com/dubinc/dub/assets/28986134/bbf2bab1-a1a0-4f77-912d-a535fa0052a1


packages/ui/src/filter/filter-list.tsx
```diff
@@ -33,14 +33,27 @@ export function FilterList({
...
               if (!filter) {
                 throw new Error(
                   "Filter.List received an activeFilter without a corresponding filter",
                 );
               }
 
-              const option = filter.options?.find((o) => o.value === value);
+              const option = filter.options?.find(
+                (o) => o.value.toLowerCase() === value.toLowerCase(),
+              );
```
TWilson023: @steven-tey what did you run into that required this change? I think it's breaking things for non-string values

steven-tey: basically the LLM will sometimes generate stuff in lowercase, which is why we need to make the comparison in lowercase to account for that – perhaps there's a better alternative here?

TWilson023: Restored this, just with extra `typeof` checks 👍

TWilson023 (approved):

PR #974, opened 2024-06-07
Quick doc fix

versecafe:
TB CLI has a max python version of 3.11 so currently breaks when using an up to date python install so added the cap as otherwise the install fails and throws a nice pile of error messages.

> also there's a random change being registered on line 361 and 459 in the files but I have no idea what maybe a whitespace change from something but can be ignored


steven-tey:
we moved our docs to https://github.com/dubinc/docs – would you be able to create this PR there instead? thanks!

PR #1093, opened 2024-07-21
Added title and description to the link card 

### LINKED ISSUE #682: feat - view or add title/description in the individual short link card

mohammedyunus2002:
Closes #682 
Screenshot 2024-07-21 at 3 46 19 PM


steven-tey:
This has been shipped via #1021

PR #1102, opened 2024-07-24
fix: correct the isBlacklistedReferrer logic

anyuj:
fixed logic for blacklisted referrer


steven-tey:
the current logic is correct

PR #1109, opened 2024-07-26
ENG-241: CSV Link Imports

### LINKED ISSUE #957: import from Shlink
https://shlink.io/

Shlink also exports to CSV so importing CSV would work too.

TWilson023:
Closes #957 as well


apps/web/lib/api/links/process-link.ts
```diff
@@ -374,7 +374,6 @@ export async function processLink<T extends Record<string, any>>({
       key,
       // we're redefining these fields because they're processed in the function
       url,
-      createdAt: createdAt ? new Date(createdAt) : undefined,
```
TWilson023: @steven-tey I think we still need to pass `createdAt` here so that the processed link retains it from the original payload

apps/web/lib/zod/schemas/import-csv.ts
```diff
@@ -1,11 +1,12 @@
...
 
 export const linkMappingSchema = z.object({
   domain: z.string(),
   url: z.string(),
   key: z.string(),
-  createdAt: z.string().optional(),
-  tags: z.string().optional(),
-  title: z.string().optional(),
-  description: z.string().optional(),
+  createdAt: parseDateSchema.nullish(),
```
TWilson023: I think this should still just be a string - it's the mapping to a CSV column header, not a data value

steven-tey: yup

apps/web/lib/api/links/process-link.ts
steven-tey: so I think we're not supposed to be able to pass `createdAt` since it should be the date the link gets created in Dub. the only exception is in link imports, so we'll need to handle that separately

TWilson023: Okay I guess we can set it after calling `processLink` or something

apps/web/app/api/cron/import/csv/route.ts
```diff
@@ -26,23 +26,19 @@ export async function POST(req: Request) {
...
+        domain: getPrettyUrl(row[mapping.domain]).split("/")[0],
+        // domain.com/path/to/page => path/to/page
         key:
-          row[mapping.key]
-            ?.replace(/^https?:\/\//, "")
-            .split("/")
-            .at(-1) ?? "",
-        createdAt:
-          createdAt && !isNaN(createdAt.getTime()) ? createdAt : undefined,
+          getPrettyUrl(row[mapping.key]).split("/").slice(1).join("/") ??
```
TWilson023: I don't think this will work if the mapped `key` is actually just a key and not the full shortlink.

e.g. `"abcde".split("/").slice(1).join("/")` is `""` instead of `"abcde"`

devkiran (approved):

steven-tey (approved):

PR #1131, opened 2024-08-02
ENG-480: Mobile Improvements

TWilson023:


apps/web/ui/analytics/top-links.tsx
```diff
@@ -43,7 +43,9 @@ export default function TopLinks() {
                         className="size-5 sm:size-5"
                       />
                     ),
-                    title: !showUrls && d["shortLink"] ? d["shortLink"] : d.url,
+                    title:
+                      (!showUrls && d["shortLink"] ? d["shortLink"] : d.url) ??
+                      "Unknown",
```
TWilson023, line 48: I don't think this will affect production, but I was running into some errors when my local data didn't have some links that Tinybird had data for

apps/web/ui/links/link-card.tsx
```diff
@@ -14,7 +16,7 @@ export function LinkCard({ link }: { link: ResponseLink }) {
       <AddEditLinkModal />
       <CardList.Card
         key={link.id}
-        onClick={() => setShowAddEditLinkModal(true)}
+        onClick={isMobile ? undefined : () => setShowAddEditLinkModal(true)}
```
TWilson023, line 19: Disabling clicking on the entire card for mobile - we could probably do something similar to the `modal-backdrop` checks instead, but I feel like with some of the small touch targets we have on mobile, this removes more potential frustration.

apps/web/ui/modals/add-edit-link-modal/index.tsx
```diff
@@ -363,7 +363,7 @@ function AddEditLinkModal({
         }
       }}
     >
-      <div className="scrollbar-hide grid max-h-[95vh] w-full divide-x divide-gray-100 overflow-auto md:grid-cols-2 md:overflow-hidden">
+      <div className="scrollbar-hide grid max-h-[95dvh] w-full divide-x divide-gray-100 overflow-auto md:grid-cols-2 md:overflow-hidden">
```
TWilson023, line 366: `dvh` seems to do a better job here so that the max height is consistent relative to the actual viewport - we should try to test this out some more in different devices/browsers.

packages/ui/src/modal.tsx
```diff
@@ -57,7 +57,7 @@ export function Modal({
         <Drawer.Portal>
           <Drawer.Content
             className={cn(
-              "fixed bottom-0 left-0 right-0 z-50 mt-24 rounded-t-[10px] border-t border-gray-200 bg-white",
+              "fixed bottom-0 left-0 right-0 z-50 rounded-t-[10px] border-t border-gray-200 bg-white",
```
TWilson023, line 60: This was preventing the link modal from filling up the available space on mobile, and I'm not seeing any issues from removing it.

apps/web/ui/analytics/top-links.tsx
steven-tey: oh gotcha, yeah good catch, let's go with this just in case 👍

apps/web/ui/links/link-card.tsx
steven-tey: okay sounds good!

apps/web/ui/modals/add-edit-link-modal/index.tsx
steven-tey: good idea!

packages/ui/src/modal.tsx
steven-tey: AMAZING!

steven-tey (approved):

PR #1147, opened 2024-08-05
ENG-475: `InputSelect` improvements

TWilson023:
Adds collision awareness (decreases height to a certain point, then flips)
Screenshot 2024-08-05 at 4 38 21 PM

Shows all values initially
Screenshot 2024-08-05 at 4 37 55 PM


steven-tey (approved):
LGTM LFG!

PR #1202, opened 2024-08-19
Add support for underscore in short links

steven-tey:


apps/web/app/api/cron/import/csv/route.ts
```diff
@@ -133,7 +133,8 @@ export async function POST(req: Request) {
           // Find tags that need to be added to the workspace
           const tagsNotInWorkspace = selectedTags.filter(
             (tag) =>
-              !tags.find((t) => t.name === tag) && !addedTags.includes(tag),
+              !tags.find((t) => t.name.toLowerCase() === tag.toLowerCase()) &&
+              !addedTags.find((t) => t.toLowerCase() === tag.toLowerCase()),
```
TWilson023, line 145: Is this safe to do without updating `processLink` to treat `tagNames` as case-insensitive?

TWilson023: Oh maybe it is, looks like the `name` column is case-insensitive in the database

TWilson023 (approved):

PR #1208, opened 2024-08-21
ENG-520: Dub Referrals

TWilson023:


packages/tinybird/pipes/v1_events.pipe
```diff
@@ -124,6 +125,7 @@ SQL >
         wl.domain as domain,
         wl.key as key,
         dse.url as url,
+        dse.continent as continent,
```
TWilson023, line 128: @steven-tey I had to add these to fix schema parse errors when loading leads or sales locally. Do you know what's different in production that's making it work there?

steven-tey: ah yes good catch – we did add this in production but didn't get around to updating these pipes, thank you!

steven-tey (approved):

PR #1210, opened 2024-08-22
ENG-520: Dub Referrals (Continued)

TWilson023:


packages/blocks/src/gauge.tsx
```diff
@@ -0,0 +1,32 @@
+import { PropsWithChildren } from "react";
+
+export type GaugeProps = PropsWithChildren<{
+  value: number;
+  min?: number;
+  max: number;
+}>;
+
+export function Gauge({ value, min = 0, max, children }: GaugeProps) {
```
TWilson023, line 9: Not a great Gauge implementation, just threw this together in a few minutes - we should probably do this properly with visx at some point.

steven-tey: this works for now, but agreed!

steven-tey (approved):

steven-tey (approved):

PR #1284, opened 2024-09-24
Add smart truncate algorithm

steven-tey:
don't like this implementation, hence a draft for now


lyon93:
I think you can use **tldts** package, and implement something like this 
`
const truncateDomain = (domain: string, maxLength: number): string => {
  const { publicSuffix: tld, domainWithoutSuffix: rest } = parse(domain);

  const minRestLength = 3;

  if (domain.length <= maxLength) {
    return `${rest}.${tld}`;
  }

  if (maxLength <= minRestLength + tld.length + 1) {
    return `${rest.slice(0, minRestLength)}.${tld}`;
  }

  const truncatedRest = truncate(rest, maxLength - tld.length - 1)!;
  return `${truncatedRest}${rest.slice(-2)}.${tld}`;
};`

steven-tey:
@lyon93 great idea, will look into it 👀

PR #1296, opened 2024-09-29
fix(ui): handle non-domain titles in DomainCardTitleColumn component

BharathxD:
[Bug Video](https://github.com/user-attachments/assets/f90ea944-b3c1-4991-8ad7-06a98163b7d0)

This pull request addresses an issue in the `DomainCardTitleColumn` component where titles passed as props were incorrectly assumed to be domains. When non-domain titles were provided, the component attempted to render them as links to `http://<title>`, leading to incorrect behavior.

#### Changes

- **Updated `DomainCardTitleColumn` component:**
  - **Rendering Logic:**
    - Utilized the `isValidUrl` function to check if the `title` is a valid URL.
    - If `title` is a valid URL, render it as a clickable link.
    - If not, render `title` as plain text within a `<div>` element, maintaining consistent styling.

#### Before and After

**Before Fix:**

- Non-domain titles were rendered as links pointing to `http://<title>`, which was incorrect and could lead to broken links.

**After Fix:**

- Non-domain titles are rendered correctly as plain text.
- Titles that are valid URLs continue to be rendered as clickable links.

### DIFF
--- a/apps/web/ui/domains/domain-card-title-column.tsx
+++ b/apps/web/ui/domains/domain-card-title-column.tsx
@@ -46,7 +46,6 @@ export function DomainCardTitleColumn({
           ) : (
             <div
               className="truncate text-sm font-medium"
-              title={domain}
             >
               {domain}
             </div>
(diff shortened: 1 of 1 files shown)


steven-tey (approved):
Great catch, thank you!

PR #1303, opened 2024-10-02
Add async combobox for link builder domains

TWilson023:


apps/web/lib/swr/use-domains.ts
```diff
@@ -36,24 +47,36 @@ export default function useDomains({
...
+    ],
+    [allWorkspaceDomains, workspaceId],
+  );
+  const allActiveDomains = useMemo(
+    () => [
+      ...(activeWorkspaceDomains || []),
+      ...(workspaceId === `ws_${DUB_WORKSPACE_ID}` ? [] : activeDefaultDomains),
+    ],
+    [activeWorkspaceDomains, activeDefaultDomains, workspaceId],
+  );
```
TWilson023, line 79: Added a bunch of `useMemo`s to prevent these from updating every render

apps/web/ui/links/short-link-input-new.tsx
```diff
@@ -439,3 +403,74 @@ function DefaultDomainPrompt({
...
+}: {
+  domain?: string;
+  setDomain: (domain: string) => void;
+  loading: boolean;
+}) {
+  const [search, setSearch] = useState("");
+  const [debouncedSearch] = useDebounce(search, 500);
+  const { domains, loading: loadingDomains } = useAvailableDomains({
+    search: debouncedSearch,
+  });
```
TWilson023: No need for passing a `currentDomain` anymore because the Combobox handles displaying the current selection even if it's not in the `options` array

steven-tey (approved):

PR #1308, opened 2024-10-02
ENG-597: Sidebar layout

TWilson023:


packages/ui/src/max-width-wrapper.tsx
```diff
@@ -10,10 +10,7 @@ export function MaxWidthWrapper({
 }) {
   return (
     <div
-      className={cn(
-        "mx-auto w-full max-w-screen-xl px-2.5 lg:px-20",
-        className,
-      )}
+      className={cn("mx-auto w-full max-w-screen-xl px-3 lg:px-10", className)}
```
TWilson023, line 13: @steven-tey I think this `MaxWidthWrapper` component is used in the marketing site so we'll need to double check / tweak things on that end.

steven-tey: ooooh yes 👀 I can publish a new version for `@dub/ui` to check

steven-tey (approved):

Issue #465, opened 2023-11-13
Short link Exact Match Not Accounted for in Search Filter


abass:
Sometimes I sanity check a short link in Dub (e.g. we hard-coded a URL in a video clip), so writing it out how we sees it in the video clip that our editor sent over, and then searching it in Dub, I would sorta expect it to show if you filter the entire URL (essentially, it's an exact match of the short link), but then I need to edit it to remove the URL. Then to ensure that I typed the short link correctly, I then use CMD + F.

Yes I can also just type in the URL and enter but I prefer to not get a click from that especially when I can do it without doing that by checking Dub.

Using site search for Dub:
![CleanShot 2023-11-13 at 10 43 12](https://github.com/steven-tey/dub/assets/5505994/a83b20e3-8db0-4bbb-8f3f-81fbbb76009e)

Dub's result:
![CleanShot 2023-11-13 at 10 41 33](https://github.com/steven-tey/dub/assets/5505994/79e62583-852a-4db5-9062-4e52dfdbf63e)

When removing the domain:
![CleanShot 2023-11-13 at 10 41 48](https://github.com/steven-tey/dub/assets/5505994/2ac4655f-6542-466b-a360-65e52944e4ab)

steven-tey:
In the spirit of Hacktoberfest – would love to finally fix this once and for all. 

This will be split into 2 PRs:
1. Add a new column in the links table called `shortLink` with a unique index
2. Make sure we update the `shortLink` field in `createLink`, `updateLink`, `bulkCreateLinks`, `bulkUpdateLinks`
3. Remove the current `shortLink` polyfill that we're doing in `transformLink`
4. Add a script that backfills the `shortLink` field for existing links (`where: { shortLink: null }`)
5. Ship the PR to production first
6. Run the backfill script
7. In a follow-up PR, update the search logic to use the `shortLink` field instead `key`
8. Ship the second PR to production

PR #1309, opened 2024-10-02
fix: added shortLink in links table with backfill script and create update feature including bulk

01vishu:
Fixes: https://github.com/dubinc/dub/issues/465

1. Added a new column in the links table called `shortLink` with a unique index
2. Updated the `shortLink `field in `createLink`, `updateLink`, `bulkCreateLinks`, `bulkUpdateLinks`
3. Removed the current `shortLink `polyfill that we're doing in `transformLink`
4. Added a script that backfills the `shortLink `field for existing links `(where: { shortLink: null })`


apps/web/lib/api/links/utils/transform-link.ts
```diff
@@ -9,25 +8,11 @@ export type LinkWithTags = Link & {
...
-    },
-  });
-
   return {
     ...link,
-    shortLink,
     tagId: tags?.[0]?.id ?? null, // backwards compatibility
     tags,
-    qrCode: `https://api.dub.co/qr?url=${qrLink}`,
+    qrCode: `https://api.dub.co/qr?url=${link.shortLink}`,
```
devkiran: `qrLink` needs `?qr=1` to identify the scan from the QR code.

apps/web/lib/api/links/bulk-update-links.ts
```diff
@@ -35,11 +41,17 @@ export async function bulkUpdateLinks(
...
-          id: linkId,
-        },
+    linkIds.map(async (linkId) => {
+      const link = await prisma.link.findUnique({
+        where: { id: linkId },
+      });
+
+      if (!link) {
+        throw new Error(`Link with id ${linkId} not found`);
+      }
```
devkiran: We're validating the link on the route level, so this additional fetch is unnecessary. Can we remove this?

devkiran (changes requested):
@01vishu Thanks for this PR! Requesting some minor changes.

apps/web/lib/api/links/utils/transform-link.ts
01vishu: Sure I'll add this param

apps/web/lib/api/links/bulk-update-links.ts
01vishu: I was facing some issue that's why i used this approach. Let me check again then I'll connect to you

01vishu: Yes, we need to fetch link because bulkUpdateLinksBodySchema contains data as boolean

I'm updateing `?qr=1` in `qrCode` link

if still there is any issue regarding to this fetch please tell me

devkiran: @01vishu I just realised we don't support updating `key` and `domain` via `/api/links/bulk`. So there is no need to update the `shortLink` I guess.

01vishu: @devkiran should we fix link search using key and domain only or keep shortLink and use it for search

What is your opinion on this?

steven-tey: > I just realised we don't support updating key and domain via /api/links/bulk. So there is no need to update the shortLink I guess.

yup exactly, will remove the changes here

steven-tey (approved):

devkiran (approved):

PR #1310, opened 2024-10-03
New Link Builder GA

steven-tey:


TWilson023 (approved):
Haven't tested myself but the code changes look right!

PR #1312, opened 2024-10-03
Optimize disposable email domain validation using `SISMEMBER` instead of `SMEMBERS`

### LINKED ISSUE #1304: Optimize disposable email domain validation using `SISMEMBER` instead of `SMEMBERS`
Currently, the validation logic for checking if an email belongs to a disposable domain uses `SMEMBERS` to retrieve all the domains from the Redis set `disposableEmailDomains`, followed by an `includes()` check in the application code. This approach can be inefficient, especially as the size of the set grows, since both the Redis `SMEMBERS` command and the subsequent membership check have an overall time complexity of **O(N)**.

https://github.com/dubinc/dub/blob/628eb46bfe7012ac3f7d316e0584265b2d999c00/apps/web/lib/actions/create-user-account.ts#L33-L42

**Suggested Improvement:** To optimize this, you can directly use Redis’s `SISMEMBER` command, which has a constant time complexity **O(1)**, to check if the domain is part of the set without retrieving all the elements. 

**Proposed Code Change:**

```ts
const domain = email.split("@")[1];
const isDisposable = await redis.sismember("disposableEmailDomains", domain);

if (isDisposable) {
  throw new Error(
    "Disposable email addresses are not allowed. If you think this is a mistake, please contact us at support@dub.co",
  );
}
```

Another benefit the app can get from the proposed solution is **lower network overhead**: you will avoid fetching the entire set from Redis, reducing the amount of data transferred between Redis and the application.

**Links:**
- https://redis.io/docs/latest/commands/smembers/
- https://redis.io/docs/latest/commands/sismember/
- https://upstash.com/docs/redis/sdks/ts/commands/set/sismember#sismember

unrenamed:
This resolves #1304

### DIFF
--- a/apps/web/lib/actions/create-user-account.ts
+++ b/apps/web/lib/actions/create-user-account.ts
@@ -1,7 +1,7 @@
 "use server";
 
 import { prisma } from "@/lib/prisma";
-import { ratelimit, redis } from "@/lib/upstash";
+import { ratelimit } from "@/lib/upstash";
 import { flattenValidationErrors } from "next-safe-action";
 import { getIP } from "../api/utils";
 import { hashPassword } from "../auth/password";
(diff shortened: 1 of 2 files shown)


devkiran (approved):

steven-tey (approved):

steven-tey:
Beautiful, thank you so much!

PR #1323, opened 2024-10-04
Create links for domains in csv imports

steven-tey:


apps/web/app/api/cron/import/csv/route.ts
```diff
@@ -182,29 +188,28 @@ export async function POST(req: Request) {
...
+                  userId,
+                  projectId: workspace.id,
+                }),
               ),
             ]);
           }
 
           addedDomains.push(...domainsNotInWorkspace);
 
           // Process all links, including domain links
```
TWilson023: Might just want to update this comment since we're not including domain links here now

TWilson023 (approved):
I could've sworn this was working before, but this change looks good anyways as long as things seem to be working in your testing :+1:

PR #1342, opened 2024-10-09
fix: warning message on duplicate destination URL

### LINKED ISSUE #926: [FEAT]: Warning popup when duplicate destination URL exists

01vishu:
fixes #926
Screenshot from 2024-10-09 22-31-35


steven-tey:
Could we use a similar design/interaction as Linear's duplicate issues UX please?

image

01vishu:
> Thanks for the PR! Could we use a similar design/interaction as Linear's duplicate issues UX please?
> 
> image

sure @steven-tey

01vishu:
> > Thanks for the PR! Could we use a similar design/interaction as Linear's duplicate issues UX please?
> > image
> 
> sure @steven-tey


https://github.com/user-attachments/assets/ba04b626-7c2e-40ca-be27-2f7e2fceeb3e



@steven-tey What do you think about this change?

apps/web/app/api/links/verify/destination-url/route.ts
```diff
@@ -0,0 +1,47 @@
+import { DubApiError } from "@/lib/api/errors";
+import { withWorkspace } from "@/lib/auth";
+import { prisma } from "@/lib/prisma";
+import { NextResponse } from "next/server";
+
+export const GET = withWorkspace(
+  async ({ req, workspace }) => {
+    const { searchParams } = new URL(req.url);
+    const url = searchParams.get("url");
```
TWilson023: I think we can just use the `searchParams` passed from `withWorkspace` here - let me know if that doesn't work!

```suggestion
  async ({ workspace, searchParams }) => {
    const { url } = searchParams;
```

apps/web/ui/links/destination-url-input.tsx
```diff
@@ -31,6 +46,31 @@ export const DestinationUrlInput = forwardRef<
...
+
+      if (debouncedUrl && debouncedUrl !== prevUrl) {
+        setPrevUrl(debouncedUrl.toString());
+        fetch(
+          `/api/links/verify/destination-url?url=${encodeURIComponent(debouncedUrl.toString())}&workspaceId=${workspaceId}`,
+        )
+          .then((res) => res.json())
+          .then((data) => {
+            setDuplicateLinks(data.exists ? data.links : []);
+          });
```
TWilson023, line 69: I think we need to add handling for **editing** a short link so that itself cannot be counted as a duplicate. We'll need to either update the `api/links/verify/destination-url` endpoint to accept a link ID to filter out, or move the `exists` determination to the client side and filter out any current link ID.

steven-tey:
We're going to implement this in-house instead, but thank you for your PR!

PR #1349, opened 2024-10-10
fixes #1348 i have fixed the scroll bar issue

love-sonkar:
fixed #1348 
i have fixed the scroll bar issue.

image


steven-tey:
this does not fix the issue

love-sonkar:
i checked on my machine it's working fine 

the navbar width is causing issue so i gave 96% width so that it won't overflow

PR #1355, opened 2024-10-10
Changing overflow on help popup to hide horizontal scroll bar

zmeyer44:
There is an unnecessary horizontal scrollbar on the help popup due to the class `overflow-scroll` so this PR changes the class to `overflow-y-scroll`
Screenshot 2024-10-10 at 1 55 51 PM
Screenshot 2024-10-10 at 1 56 38 PM


steven-tey (approved):

steven-tey:
Thank you!

PR #1612, opened 2024-10-22
Fix: modal content getting covered by Drawer Vaul

### LINKED ISSUE #1609: UI Breaks on `Add New API Key` Drawer with `Restricted` Permissions on Mobile
### Issue Summary

When accessing the `Settings > API Keys` page on mobile, opening the `Add new API Key` drawer and selecting the `Restricted` permissions option causes UI issues. The `Create API key` button becomes unclickable, and the animated list of restricted permissions appears incomplete, as it is cut off at the bottom. This issue prevents users from completing the process of adding a new API key with restricted permissions.

### Steps to Reproduce

1. Open the application on a mobile device.
2. Navigate to `Settings > API Keys`.
3. Click on the `Create` button to open the drawer.
4. Select the `Restricted` permissions option from the available options.
5. Try to scroll through the animated list of restricted permissions.

### Expected behavior

1. The UI should display the full list of restricted permissions without being cut off at the bottom.
2. The `Create API Key` button should remain clickable and functional, allowing the user to proceed with adding a new API key.

### Screenshots/Videos

https://github.com/user-attachments/assets/2d389a85-ca4a-4500-82e9-5b30dcc6a019

### Version info

Tested on both:
1. iPhone 8+ iOS 16.7.10
2. Brave 1.71.114 Chromium: 130.0.6723.58 (Official Build) (arm64)

The video was recorded using Brave browser. The issue is the same on iOS, although looks just a bit different (the scroll does not work at all, i.e. there is no empty white space at the bottom)

aritradevelops:
Fixes #1609

# Issue Description

The common [`Modal Component`](https://github.com/dubinc/dub/blob/main/packages/ui/src/modal.tsx) built on top of [Drawer from Vaul](https://www.npmjs.com/package/vaul) ships with some [default css](https://github.com/emilkowalski/vaul/blob/main/src/style.css) which is introducing an overlay with an ::after attribute on top of the modal content on smaller devices. See the below images

image

Now if you take a good look at `data-attributes` the css activates on two condition when `direction` is `bottom` and no `custom-container` is not passed, which is our case.

# Trials
1. I have tried using negative z index that makes the content visible but the after block with `200%` height is there so the layer is scrollable without any content over there. (x)
2. I have tried changing different attributes to the Drawer element used in [`Modal Component`](https://github.com/dubinc/dub/blob/main/packages/ui/src/modal.tsx) but got no luck. (x)
3. The Drawer Component does not allow to overwrite its className, it does not accept the className attribute.(x)
4. Providing Custom container gives Client Rect Boundary error.(x)

# Fix
The only way to grab the element and set its height properly was to use `css` and as every component is sharing the same modal therefore it is a global issue so added the css in global.css file.

# After the fix
Verified that it works on smaller devices as well as tabs and laptops. Find the screen shots below

image
image


aritradevelops:
Merge branch 'main' into aritradevelops:fix/modal_overwritting_content

PR #1618, opened 2024-10-23
Fix: Distorted screenshot on login page

timzolleis:
## What does this PR do?

Fixes the distortion of the dub screenshot (especially on ultrawide screns) on the app.[domain]/login page (see screenshot attached)
screenshot-dub

## How should this be tested?
- Open the login page
- Use a wide monitor / wide aspect ratio


steven-tey:
incredible, thank you!

PR #1628, opened 2024-10-24
Partner Portal

steven-tey:


apps/web/lib/auth/options.ts
```diff
@@ -415,6 +416,19 @@ export const authOptions: NextAuthOptions = {
...
+
+      // Allows callback URLs on the same origin or from additionally allowed hostnames
+      if (
+        new URL(url).origin === baseUrl ||
+        PARTNERS_HOSTNAMES.has(new URL(url).host)
+      )
+        return url;
+
+      return baseUrl;
+    },
```
TWilson023: Added this to allow redirecting to `partners.dub.co` instead of `app.dub.co` when logging out from the partners portal

Issue #1651, opened 2024-10-25
CSV download option visible to non-business users on Events page, causing unclear error


unrenamed:
### Issue Summary

On the Events page, users with a billing plan lower than Business can see the option to download a CSV of the Real-time Events Stream, but clicking it results in a vague "Error" toast message. This error message does not explain the root cause, and ideally, the CSV download option should be hidden for users on plans below Business since the feature is not accessible to them.

### Steps to Reproduce

1. Log in with an account on a billing plan below Business.
2. Navigate to the `Events` page.
3. Attempt to download a CSV of the Real-time Events Stream by clicking the download option.
4. Observe the "Error" toast message that appears without further explanation.

###  Expected Behavior

For users on plans below Business, the CSV download option ideally should be hidden, preventing access to features not included in their subscription. 

### Alternative approaches

If the option remains visible, a clear error message should appear, explaining that CSV downloads are available only for Business plans.

### Reproduction Video

https://github.com/user-attachments/assets/3253c9b2-4fa4-4c78-827d-48b81dd7c56e

steven-tey:
/award 150 great catch! Lets go ahead and disable the download csv button (not the `⋮` dropdown) when `plan === 'free'` or `plan === 'pro'`

PR #1673, opened 2024-10-28
fix(web): disable events CSV download for non-business users

unrenamed:
## What does this PR do?

- [x] disables events CSV download for non-business users

Fixes: #1651

## How to test?

For users on plans below Business, the `Events` CSV download option should be disabled and a tooltip should suggest upgrading to the Business plan.


unrenamed:
@steven-tey This one is ready for peer review.

steven-tey (approved):

PR #1702, opened 2024-11-06
Payouts

steven-tey:


apps/web/lib/actions/create-dots-transfer.ts
```diff
@@ -0,0 +1,45 @@
...
+    const [transfer, orgTransfer] = await Promise.all([
+      createTransfer({
+        amount,
+        dotsAppId: workspace.dotsAppId,
+        dotsUserId,
+      }),
+      createOrgTransfer({
+        amount: fee,
+        dotsAppId: workspace.dotsAppId,
+      }),
```
TWilson023: @steven-tey we may want to fetch this fee on the back-end rather than letting the front-end pass any amount

steven-tey: great idea, will restrict this to only pass `payoutId` and fetch the details server side

devkiran (approved):

PR #1703, opened 2024-11-07
Program lander

steven-tey:


apps/web/package.json
```diff
@@ -23,7 +23,7 @@
     "@hookform/resolvers": "^3.9.0",
     "@next-auth/prisma-adapter": "^1.0.5",
     "@planetscale/database": "^1.18.0",
-    "@prisma/adapter-planetscale": "5.18.0",
+    "@prisma/adapter-planetscale": "5.14.0",
```
TWilson023: Had to downgrade `@prisma/adapter-planetscale` to get unicode characters to work properly in the `lander` JSON - seems to be related to [this issue](https://github.com/prisma/prisma/issues/24592).

Not sure if this will cause us any other problems...

steven-tey: sent feedback in slack! feel free to resolve this once you change to just `prisma`

apps/web/ui/partners/lander-blocks/ImageBlock.tsx
```diff
@@ -0,0 +1,18 @@
...
+}: {
+  block: z.infer<typeof programLanderImageBlockSchema>;
+}) {
+  return (
+    <div className="py-4">
+      <img
+        src={block.data.url}
+        alt={block.data.alt}
+        className="block rounded-lg"
+      />
```
TWilson023, line 15: Should we store image size / aspect ratio data so we can avoid layout shifts when images load?

devkiran (approved):

PR #1713, opened 2024-11-13
ensure 'www.' is only removed at the start of domain

mkreuzmayr:
While browsing the repo, I stumbled upon a potential issue in how the domain was being sanitized. The current implementation removes `www.` from any part of the domain string, which could lead to incorrect behavior imo.
This PR fixes that by ensuring `www.` is only removed if it appears at the beginning of the domain.

`example.www.com` → `example.com` (incorrectly removes 'www' from the middle)


steven-tey:
Great catch!! Thank you 🙏 

cc @devkiran to review

steven-tey (approved):

devkiran (approved):

Issue #1732, opened 2024-11-18
QR codes without margin are hard to scan


zhengkyl:
I saw one of your QR codes in the wild, and after digging I noticed that your generators have margins disabled by default.

![image](https://github.com/user-attachments/assets/a2faa524-fbe5-4989-bc71-53e9df3533a7)

This is fine if the background is "white", but they won't work well on dark backgrounds.
| Easy to scan | Slow/Difficult to scan |
|--------|--------|
| ![with](https://github.com/user-attachments/assets/4db498d1-f089-4d96-a72c-70b821033333)| ![without](https://github.com/user-attachments/assets/e14bffef-2199-4b2b-a2f3-c9ba141ca522) |

![image](https://github.com/user-attachments/assets/59a0708f-6b7e-4608-99da-b6e5e948f08a)

The 3 large squares are used to "find" the QR code, and the edges need to stand out. The margin doesn't need to be large, 1 module wide is usually fine, 2 wide is safe. 4 wide is unnecessary. https://github.com/dubinc/dub/blob/c5db6dcc271fb539f2f8068dcc2600ebf5a82ff6/apps/web/lib/qr/constants.ts#L16

steven-tey:
Fixed by #1733 – thanks again for the feedback & the PR @zhengkyl!

PR #1733, opened 2024-11-18
Include margin on QR codes by default

zhengkyl:
Addresses #1732 and also changes default margin size.

This would be a change to the API and also slightly affect the preview UI, so perhaps other changes are needed like a margin toggle button for anyone who wants the current behavior.


steven-tey:
Thanks a lot for the PR @zhengkyl! Just merged this 🙏

Issue #1810, opened 2024-12-19
Removing a tag while editing an existing link breaks the website


hkirat:
I tried editing an existing link. I removed a tag from it (there was only 1 tag in there)
As soon as I unselected the tag, the website crashed

<img width="1862" alt="Screenshot 2024-12-19 at 5 05 51 PM" src="https://github.com/user-attachments/assets/a231a82a-9652-4526-838b-44ce30ec9d03" />

hkirat:
Cant re-produce it reliably

steven-tey:
thanks for the report, fixing!

PR #1813, opened 2024-12-19
fix(tag-selection): fix tag deselection crash in link editing

ThisChirag:
## Problem
The issue #1810 was marked as resolved but i'm still facing the same issue:
"Removing the only tag during link editing still causes the website to crash."

## Solution
This fix ensures:
- Prevention of invalid states when all tags are removed.

## Testing
- Tested removing all tags and re-adding them.

References #1810


steven-tey:
Thanks for the PR! Can you outline specific steps for reproducing the original error?

ThisChirag:
> Thanks for the PR! Can you outline specific steps for reproducing the original error?

Thank you for reviewing the PR. Here's how to reproduce the original error:

- Open the edit link modal for any existing link that already has at least one tag assigned.
- Remove all assigned tags by deselecting them, or deselect the only assigned tag.
- After deselecting all the tags, the website crashes.

**Expected Behavior:**
- The tags should be successfully removed.
- The UI should remain functional, and changes should be saved without issues.

**Observed Behavior:**
- The system crashes immediately after deselecting all the tags, causing page to become unresponsive.
- The crash occurs because the system attempts to process an empty tag state, which results in invalid behavior.

This fix ensures that invalid states (such as an empty tag state) are handled properly, preventing the crash.

Let me know if any specific details are required.
Thank You.

steven-tey:
This should be fixed now 🙏 but let us know if you're still running into this issue

PR #1821, opened 2024-12-21
Fix truncated focus

michaltarasiuk:
I removed the overflow-hidden class from the AnimatedSizeContainer component. After reviewing its usage, I did not find any display issues.

Before:
Screenshot 2024-12-21 at 21 49 18
After:
Screenshot 2024-12-21 at 21 50 38

### DIFF
--- a/packages/ui/src/animated-size-container.tsx
+++ b/packages/ui/src/animated-size-container.tsx
@@ -38,7 +38,6 @@ const AnimatedSizeContainer = forwardRef<
     return (
       <motion.div
         ref={forwardedRef}
-        className={cn("overflow-hidden", className)}
         animate={{
           width: width
             ? resizeObserverEntry?.contentRect?.width ?? "auto"


steven-tey:
Sorry I don't see the issue here – what's wrong with the current input otp implementation?

michaltarasiuk:
The issue with the current implementation is that the focus indicator is partially truncated when the user interacts with the OTP input field. This happens because the overflow-hidden class in the AnimatedSizeContainer prevents elements that extend beyond the container’s boundaries from being fully visible.

By removing the overflow-hidden class, the focus indicator is now fully visible, which resolves this problem.

@steven-tey Let me know if this clarifies the issue!

TWilson023 (changes requested):
I don't think we'll be able to remove `overflow-hidden` from `AnimatedSizeContainer`, because we don't want content to overflow while the container is still expanding.

Here's what happens when we remove the `overflow-hidden` (slowed down to help demonstrate):

https://github.com/user-attachments/assets/206cf304-7afa-4194-bfec-72962645fd80

Instead, we may just need to play around with paddings/margins to get this focus ring to fit within the `overflow-hidden` container.

Issue #1780, opened 2024-12-08
bug: can't exit search bar using ESC key (to use keyboard shortcuts again)


daallgeier:
1. I was searching for a link using the search bar but could not find it.  
2. Thus i wanted to create a new one.  
3. I went to press ESC and `c`

### Expected
I expected the new link modal to show up

### Actual Result
I searched for `C`

steven-tey:
@daallgeier ha! good one – will get this fixed, thank you 🙏

steven-tey:
fixed by #1824, h/t @michaltarasiuk

PR #1824, opened 2024-12-21
Unable to exit the search bar using the ESC key

michaltarasiuk:
Add Escape key functionality to blur the search input, fixing issue #1780.


steven-tey (approved):

steven-tey:
perfect, thanks for the neat fix! 🙏

PR #1843, opened 2024-12-28
Update README.md with proper href path to LICENSE.md

aadamw:


steven-tey (approved):

steven-tey:
Great catch, thank you!

PR #1888, opened 2025-01-13
Fix `allowedHostnames` implementation

steven-tey:
Improve `allowedHostnames` implementation by updating the logic to check both the apex domain and the domain without the "www" prefix.


TWilson023 (approved):

codewithshek:
Great job on improving the allowedHostnames implementation! This update will definitely enhance domain validation. Thanks for the hard work!

PR #1899, opened 2025-01-15
fix: `/banned` Nav and Foooter links go to correct domain

### LINKED ISSUE #1898: Links from /banned route don't go to the correct destination
On pages like [`https://dub.sh/banned`](https://dub.sh/banned), clicking links in the nav/footer will go to `https://dub.sh/brand`, and will either end up on [`https://dub.co`](https://dub.co) or `https://app.dub.co/<workspace>` depending on if the user is logged in.

This is seemingly because the links in the `Nav` use domain-relative links, which works out on dub.co, but fails when these UI components are displayed on other domains.

I think that the solution is to prepend the domain to the links.

6TELOIV:
Fixes #1898

`/banned` is unique like [`/app.dub.co/(share)/share/[dashboardId]/page.tsx`](https://github.com/dubinc/dub/blob/main/apps/web/app/app.dub.co/(share)/share/%5BdashboardId%5D/page.tsx) in that it is on a specific domain but does not have a `[domain]` in its path. This means that the `Nav` and `NavMobile` will assume it's on `dub.co` and use links relative to the current domain, when in reality it needs to specify `dub.co`.

This is fixed by manually specifying that the domain is not `dub.co` and instead is `dub.sh` with `staticDomain="dub.sh"`


steven-tey (approved):

steven-tey:
@6TELOIV fantastic analysis and great fix, thank you! Just merged!

Issue #1894, opened 2025-01-14
[BUG] password complexity check when registering


bogusbd:
Hello, 

Password complexity is not checked when registering, but when entering OTP it is checked and I get an error:

```
0:["$@1",["LcysgsUh_M6A-tjDh0BnZ",null]]
1:{"validationErrors":{"password":["Password must contain at least one number, one uppercase, and one lowercase letter"]}}
``` 

visible in API request but not in interface

This error is reproducible on dub.co.
Perhaps it was fixed in the main branch.

Dyuti01:
Hey, I want to work on this bug, please assign me this task

Dyuti01:
#1904

steven-tey:
Fantastic fix, @Dyuti01, thank you! will close this issue for now and get your PR merged soon 👍

PR #1904, opened 2025-01-15
Fix: applying zod validation of password before sending OTP in signup.

Dyuti01:
Fixes issue: #1894 
### Reason 
The bug is because during signup only the "email" is sent as payload and not the password so zod is not finding any password to validate.

### Solution
Fixed bug in `signup` by adding `zod validation of password` before sending OTP to the user.
Also added zod-invalid email and password messages to show in toast.


Dyuti01:
Please review this pr.

steven-tey:
Great work @Dyuti01, this LGTM! @devkiran can you review and get this merged?

Dyuti01:
@devkiran 
Hello Sir, can you please review this PR since the issue was closed already.

steven-tey (approved):

steven-tey:
This looks great, thank you so much for the PR!

PR #1918, opened 2025-01-18
refactor: use next-safe-action `isPending` instead of `isExecuting`

kevinmitch14:
Using `isPending` is the recommended approach as per next-safe-action docs [(link)](https://next-safe-action.dev/docs/execute-actions/hooks/usestateaction#usestateaction-return-object): 
> For checking the action status, the recommended way is to use the isPending shorthand property. Using isExecuting or checking if status is "executing" could cause race conditions when using navigation functions, such as redirect.


steven-tey (approved):

steven-tey:
Looks awesome, thank you so much for the PR!

PR #1932, opened 2025-01-22
Persist link display settings to workspace

TWilson023:


apps/web/ui/links/links-display-provider.tsx
```diff
@@ -134,94 +131,72 @@ const parseDisplayProperties = (displayPropertiesRaw: string[]) =>
...
+        sortBy: parseSortBy(object.sortBy),
+        showArchived: parseShowArchived(object.showArchived),
+        displayProperties: parseDisplayProperties(object.displayProperties),
+      }
+    : undefined;
+
 export function LinksDisplayProvider({ children }: PropsWithChildren) {
-  const searchParams = useSearchParams();
-  const sortRaw = searchParams?.get("sortBy");
-  const showArchivedRaw = searchParams?.get("showArchived");
```
TWilson023, line 145: I think having these search params override the persisted value is just from when we first added the new filtering/persistence. If so, these should be safe enough to remove now.

steven-tey (approved):

PR #1934, opened 2025-01-22
Update signature.ts

thebedroomprogrammer:
Remove unnecessary check for the secret key as the secret is a required parameter in the function definition.


devkiran:
@thebedroomprogrammer May I know why you think it is unnecessary?

thebedroomprogrammer:
> @thebedroomprogrammer May I know why you think it is unnecessary?

Correct me if I am wrong but isn't the secret an always required parameter in the function which will fail to compile for any function that does not pass the secret ensuring that it is always there. The only time we would need to check whether the secret is present is if someone fails to pass it in the runtime.

Issue #1939, opened 2025-01-25
Bug: Upsert/Create Link API Schema questions


Flo4604:
The https://dub.co/docs/api-reference/endpoint/upsert-a-link docs say that the `publicStats` parameter is deprecated and that we should use `dashboard` instead, this parameter is not included in the openapi spec tho and looking at the createLink function I didn't find it either.

Is the dashboard parameter just yet to be added? 

----

The response of the upsert / create link includes`"identifier":null` which breaks the generated sdks as they verify the response schema.
For example:`"error unmarshalling json response body: unknown fields: [identifier]"` from the go SDK.

Can the identifier: https://github.com/dubinc/dub/blob/81ee24b33b354e7e3c7706dc0279bc669f6872e3/apps/web/lib/api/links/utils/transform-link.ts#L21 be removed from the response since its not in the schema anymore or added back to the schema with a deprecated flag?

Happy to open a PR once I know what the correct solutions would be

PR #1962, opened 2025-02-01
fix: fully deprecate link identifier

Flo4604:
This fixes the identifier issue in #1939 and makes the go-sdk usable again. 

The identifier was already removed from the openapi schema thus I opted to just fully remove it instead of backfilling it with null.


steven-tey:
@Flo4604 this actually looks like a bug in the generated Go SDK so we shouldn't need to change anything here – will get that fixed soon! Sorry about that!

PR #1995, opened 2025-02-09
Improve support for `interval='all'` with `dataAvailableFrom`

steven-tey:
Fixes https://linear.app/dubinc/issue/ENG-1035/improve-all-time-analytics-filter


TWilson023 (commented):
I might be misunderstanding things but instead of adding/passing `dataAvailableFrom` around in a ton of different places, could we just determine the `start` on the front-end based on the workspace's `createdAt`?

PR #2011, opened 2025-02-11
Program Overview Updates

TWilson023:
Optional further improvements:
- [x] Add "x of n" results to Payouts card - would just require some endpoint updates to get total counts for a specific program
- [ ] Payouts card shimmer dots hover states - has performance implications, might take some time to optimize


apps/web/app/api/partner-profile/programs/[programId]/earnings/timeseries/route.ts
```diff
@@ -5,9 +5,8 @@ import { sqlGranularityMap } from "@/lib/planetscale/granularity";
 import { analyticsQuerySchema } from "@/lib/zod/schemas/analytics";
 import { prisma } from "@dub/prisma";
 import { Prisma } from "@prisma/client";
-import { format } from "date-fns";
+import { DateTime } from "luxon";
```
Comment (TWilson023) [inline apps/web/app/api/partner-profile/programs/[programId]/earnings/timeseries/route.ts:8]: Ended up adding Luxon for proper timezone handling.

I tried using `date-fns` with `date-fns-tz` to convert to/from the specified timezone, but this couldn't handle DST properly because the functions used in `dateIncrement` still operate in the server timezone (UTC)

Comment (steven-tey) [reply apps/web/app/api/partner-profile/programs/[programId]/earnings/timeseries/route.ts:8]: nice! we should take some time soon to standardize all these across our codebase btw 🙏

steven-tey (approved):

steven-tey (approved):

PR #2113, opened 2025-03-03
fix reward import

devkiran:


steven-tey (approved):

apps/web/lib/rewardful/import-campaign.ts
```diff
@@ -25,34 +26,51 @@ export async function importCampaign({ programId }: { programId: string }) {
...
+
+    rewardId = reward.id;
   }
 
   return await rewardfulImporter.queue({
     programId,
     // we will only need to assign rewardId to affiliates
     // if it's not the defaultRewardId of the program (there's already a defaultRewardId)
-    ...(defaultRewardId && { rewardId: reward.id }),
+    ...(defaultRewardId && rewardId && { rewardId }),
```
steven-tey: Even if the reward exists (`rewardFound`), don't we still need to pass the rewardId to the users if it's not the `defaultRewardId` so it can be assigned to the partners?

Also, how would program-wide click/lead rewards work?

steven-tey (approved):

PR #2211, opened 2025-03-26
Email updates

marcusljf:
- Updates to language where we could improve
- Updated styling to the buttons to make the marketing site and app
- Update to the `new-sale-alert-program-owner` to make it more clear who's getting the sale, and where it came from.

### DIFF
--- a/packages/email/src/templates/new-sale-alert-program-owner.tsx
+++ b/packages/email/src/templates/new-sale-alert-program-owner.tsx
@@ -106,7 +106,7 @@ export function NewSaleAlertProgramOwner({
   return (
     <Html>
       <Head />
-      <Preview>You received a payment from a customer referred! 💰</Preview>
+      <Preview>You received a sale from a customer referred! 💰</Preview>
       <Tailwind>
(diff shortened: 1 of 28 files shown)


packages/email/src/templates/new-sale-alert-program-owner.tsx
```diff
@@ -123,7 +123,7 @@ export function NewSaleAlertProgramOwner({
             </Heading>
 
             <Text className="text-sm leading-6 text-neutral-600">
-              You received a payment from a customer referred by{" "}
+              <strong>{workspace.name}</strong> earned a sale from a customer referred by {" "}
```
devkiran: @marcusljf maybe use the `{program.name}` instead?

steven-tey: great call

steven-tey (approved):

packages/email/src/templates/new-sale-alert-program-owner.tsx
marcusljf: Works for me! Thanks

PR #2298, opened 2025-04-13
Fixed the bug where you could delete a workspace without entering the slug

BharathxD:
Required slug input to prevent workspace deletion without proper validation. Also cleaned up unnecessary dependencies in useCallback hook.

Good to be back! Finally got a new machine that can actually handle this beast of a codebase. Before I was trying to fix issues without even being able to run it locally, which was a real struggle 🤓


steven-tey (approved):

steven-tey:
Oh, great catch, thank you! 🙏

PR #2300, opened 2025-04-14
Switch to `customerExternalId` for better semantics

steven-tey:
Inspired by [Polar](https://docs.polar.sh/api-reference/checkouts/create-session#body-customer-external-id) :)


steven-tey:
closing for now, can revisit later

Issue #2309, opened 2025-04-16
Inaccurate click tracking due to HEAD requests from platforms like Twitter/X


BharathxD:
I noticed something weird with my link analytics that I wanted to bring to your attention. My shortened links are showing significantly more clicks than actual visitors, and I think I've figured out why.

This appears to be related to Issue #1860 about false link counts, but I've identified the specific mechanism causing the problem.

When I share links on Twitter/X, I'm seeing multiple clicks recorded instantly, even though nobody has opened them. After investigating, I found that X sends one GET request (from TwitterBot) followed by three HEAD requests for each shared link. These HEAD requests are being counted as actual clicks, inflating my analytics.

<img width="396" alt="Image" src="https://github.com/user-attachments/assets/7d43282f-412a-4177-9414-eacfd4bc6412" />

<img width="1650" alt="Image" src="https://github.com/user-attachments/assets/795b2e48-8677-4adc-b437-7446ed2368d5" />

These HEAD requests use standard browser user agents (sometimes with different IPs) rather than bot identifiers:

- iPhone Safari user agent
- Desktop Chrome user agent
- Empty user agent

I tested the same thing with Rebrandly and noticed they don't count HEAD requests as clicks, which makes sense since these are just automated platform checks, not real user interactions.

To confirm this wasn't a fluke, I:

- Shared multiple links on different platforms
- Checked server logs to verify the request patterns
- Tested with Postman using different HTTP methods

The fix seems pretty straightforward - you're already filtering out bots, but you should also filter out HEAD requests since these aren't actual user interactions but just metadata checks.

steven-tey:
Really phenomenal investigation + analysis + fix here @BharathxD, thank you so much! Excluding `HEAD` requests from being tracked makes a lot of sense – we'll get your PR reviewed and merge it soon 🙏

steven-tey:
Merged #2310 – thank you so much!

BharathxD:
Thank you, Steven! I appreciate the kind words and your time reviewing the PR.

Issue #1860, opened 2025-01-03
False Link Count 


Suksham-sharma:
Sending link via twitter , increases the count even when the link isn't opened by any person . 

It gives false analytics .

sanidhyarathore:
Can I take this up?

steven-tey:
can you give me an example of a link for this so I can repro and investigate?

Suksham-sharma:
@steven-tey , Sure , I tried to reproduce it , i generally happens , when you shorten a link and send it via , twitter / linkedin , you get a click , most prolly via bots trying to check if that link contains anything harmful or not . Which results in click getting add , even when no such action has taken place . Resulting in false analytics .

steven-tey:
We do exclude bot clicks from analytics - can you please let me know an example of a link that has this issue so I can make sure there are no bots that are missed?

steven-tey:
@Suksham-sharma We do exclude bot clicks from analytics - can you please let me know an example of a link that has this issue so I can make sure there are no bots that are missed?

Suksham-sharma:
@steven-tey , when i went through the codebase , i also noticed that bot clicks are ignored , but somehow they are being missed . 

Attaching a sample created link for reference. 

https://github.com/user-attachments/assets/8cedde15-cbea-46fb-b049-91831e72d3a0

steven-tey:
The reproduction is super helpful, thank you! Will look into it 👍

gmpaliwal07:
I guess it is working fine now cause I'm not facing False Link Count i just validated that

steven-tey:
Yup this was fixed by #2310, closing now!

PR #2310, opened 2025-04-16
Fix click tracking bug by filtering HEAD requests

BharathxD:
Addresses issue #2309 (and #1860) by ensuring only user-initiated requests are counted, preventing inflation from HEAD requests.


steven-tey (approved):

steven-tey:
Thank you so much again! Fantastic investigation + PR 👏

PR #2315, opened 2025-04-17
Revert link cache bug

steven-tey:


devkiran (approved):

apps/web/lib/api/links/cache.ts
```diff
@@ -51,7 +51,9 @@ class LinkCache {
   }
 
   async get({ domain, key }: Pick<LinkProps, "domain" | "key">) {
-    return await redis.get<RedisLinkProps>(this._createKey({ domain, key }));
+    // here we use linkcache:${domain}:${key} instead of this._createKey({ domain, key })
```
joe0xbuffer, line 54: don't we need to update the `async get` call as well to match?

PR #2333, opened 2025-04-21
Scalable daily click aggregation using batch processing

BharathxD:
This PR adds batch processing to the daily click aggregation to handle larger datasets more efficiently, addressing the TODO comment in the old code about using a cron job for scalability. Previously, processing all links at once could lead to performance issues. Now, we're using QStash to process links in batches of 100, with the GET handler initiating the first batch and the POST handler managing subsequent batches via cursor-based pagination. This approach improves scalability and reliability, reducing the risk of timeouts or resource exhaustion


devkiran:
@BharathxD, thanks for opening this PR. We're going to close this one since the cron logic has been improved using a different approach.

PR #2338, opened 2025-04-22
Webhooks: Fix infinite loading state and add Pro badge indicator

BharathxD:
### What's changed
- Fixed the webhook selector's infinite loading state by ensuring options array is never undefined
- Added Pro badge tooltip to indicate webhooks as a premium feature

### Why
The webhook selector would show an infinite loading state when `availableWebhooks` was undefined, leading to poor UX. Also, users weren't aware that webhooks are a Pro feature until they tried to use it.

Screenshot 2025-04-22 at 12 21 40 PM


steven-tey:
Appreciate the PR @BharathxD! We're actually going to be removing this soon (via an overhaul for our webhooks feature to make it more intuitive in #2276) – so I'll be closing this for now 🙏

PR #2404, opened 2025-05-08
Add icons to `@dub/ui`

TWilson023:
New icons for use on dub.co - Requires a new `@dub/ui` release


raiderss:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

steven-tey (approved):

PR #2429, opened 2025-05-21
Improve the workspace store schema

devkiran:


apps/web/lib/swr/use-workspace-store.ts
```diff
@@ -6,13 +6,13 @@ import { mutate } from "swr";
...
 import useWorkspace from "./use-workspace";
 
-export const STORE_KEYS = {
-  conversionsOnboarding: "conversionsOnboarding",
-  dotLinkOfferDismissed: "dotLinkOfferDismissed",
-};
+type STORE_KEYS =
+  | "conversionsOnboarding"
+  | "dotLinkOfferDismissed"
+  | "programOnboarding";
```
TWilson023: Can we infer this from `updateWorkspaceStoreSchema` (or a separate shared schema) so we don't need to update these keys in multiple places?

TWilson023 (commented):
Looking pretty good at a glance! Should we validate the store values as well as the keys? I did something similar for user workspace preferences (see [schemas](https://github.com/dubinc/dub/blob/main/apps/web/lib/zod/schemas/workspace-preferences.ts)), though I wonder if the implementation could be less complex

apps/web/lib/swr/use-workspace-store.ts
devkiran: good catch

devkiran: > Should we fully validate the store values in addition to the keys? 

@TWilson023 It might be a bit tricky since we store the program onboarding flow data as a value, whereas, if I'm not mistaken, the values are just booleans in most other cases.

steven-tey: yeah it might be too complex to validate `programOnboarding` json so we can leave that as is for now 👍

TWilson023 (approved):

PR #2431, opened 2025-05-21
Improve checkout.session.completed webhook

steven-tey:


steven-tey:
closing in favor of #3163

PR #2447, opened 2025-05-28
Improve payout sorting

steven-tey:
## Summary by CodeRabbit

- **New Features**
  - Added a "Paid" column to the payout table, displaying the payment date or a dash if unavailable.

- **Improvements**
  - Default sorting for the payout table is now based on the payout period end date.
  - Updated sorting and display to use "Period End" instead of "Period Start" across the payout table.
  - Enhanced sorting options to include the new "Paid" column.


steven-tey:
@coderabbitai any breaking changes?

steven-tey:
@coderabbitai we're not exposing /payouts in any public facing APIs, so I think we're fine right?

PR #2482, opened 2025-06-04
Fix customer discount implementation

steven-tey:
## Summary by CodeRabbit

- **Bug Fixes**
	- Removed the display of customer discount information from customer-related API responses. Discounts will no longer appear in customer data returned by the API.
- **Improvements**
	- Enhanced customer data to include a prioritized fallback for discount information based on program enrollment details.


steven-tey:
@coderabbit final review pls

PR #2492, opened 2025-06-05
Use default export for email templates

devkiran:
## Summary by CodeRabbit

* **Refactor**
  * Updated all email template components to use default exports instead of named exports, and adjusted import statements throughout the application accordingly. This change streamlines the way email templates are imported and exported, but does not affect any user-facing features or application behavior.


steven-tey (approved):

steven-tey:
@coderabbit final review

PR #2539, opened 2025-06-17
Don't use composite analytics by default

steven-tey:


steven-tey:
@coderabbit final review

PR #2587, opened 2025-06-27
Simplify invite partners step in onboarding flow

steven-tey:


steven-tey:
@coderabbit final review

PR #2603, opened 2025-07-03
Update ProgramInviteCard UI and added invite date

marcusljf:
- Updated the ProgramInviteCard component to be closer to the designs in styling, and remove the reward and discount display because it breaks consistency if there are multiple cards. 
- Added the invited date and wanted to create a "New" status with a counter in the sideNav, but that required backend changes (per cursor), and didn't feel comfortable going down that path.


steven-tey (approved):

steven-tey:
@coderabbit final review

PR #2627, opened 2025-07-08
Safari select fix

marcusljf:
Moved SelectionToolbar from inside the table header to an absolute overlay above the table, to make it look the same across all browsers, but mostly to fix the weird safari location.

**Before**
CleanShot 2025-07-08 at 12 57 13@2x

**After**
CleanShot 2025-07-08 at 14 30 13@2x

### DIFF
--- a/packages/ui/src/table/selection-toolbar.tsx
+++ b/packages/ui/src/table/selection-toolbar.tsx
@@ -2,13 +2,17 @@ import { cn } from "@dub/utils";
 import { Table } from "@tanstack/react-table";
 import { ReactNode, useEffect, useState } from "react";
 import { useKeyboardShortcut } from "../hooks";
+import { Checkbox } from "../checkbox";
+import { AnimatePresence, motion } from "framer-motion";
 
 export function SelectionToolbar<T>({
   table,
   controls,
+  className,
 }: {
   table: Table<T>;
   controls?: (table: Table<T>) => ReactNode;
+  className?: string;
 }) {
   const selectedCount = table.getSelectedRowModel().rows.length;
   const [lastSelectedCount, setLastSelectedCount] = useState(0);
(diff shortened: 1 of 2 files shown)


packages/ui/src/table/selection-toolbar.tsx
```diff
@@ -23,33 +27,66 @@ export function SelectionToolbar<T>({
...
   return (
-    <tr
-      className={cn("size-0")}
-      {...{
-        inert: table.getSelectedRowModel().rows.length ? undefined : "",
-      }}
-    >
-      <td colSpan={table.getHeaderGroups().length} className="contents size-0">
-        <div
+    <AnimatePresence>
```
TWilson023: Would be nice to avoid adding all of the `AnimatePresence` and `motion.*` stuff unless we really need it - CSS animations will be simpler and more performant.

packages/ui/src/table/selection-toolbar.tsx
```diff
@@ -23,33 +27,66 @@ export function SelectionToolbar<T>({
...
+      {selectedCount > 0 && (
+        <motion.div
+          initial={{ opacity: 0 }}
+          animate={{ opacity: 1 }}
+          exit={{ opacity: 0 }}
+          transition={{ duration: 0.12 }}
           className={cn(
-            "pointer-events-none absolute bottom-px left-11 right-0 top-0 bg-white opacity-0 duration-100",
-            selectedCount && "pointer-events-auto opacity-100",
+            "pointer-events-auto absolute z-10 left-0 right-0 top-0 min-h-[40px] border-b border-border-subtle bg-white",
```
TWilson023: We should avoid hardcoding this `min-h-[40px]` in multiple places, and make sure it's overridable for tables with different header row heights

packages/ui/src/table/selection-toolbar.tsx
```diff
@@ -23,33 +27,66 @@ export function SelectionToolbar<T>({
...
+            className
           )}
+          style={{ boxSizing: "border-box" }}
```
TWilson023: If we need `box-sizing: border-box` it can be in the `className` as `box-border`

packages/ui/src/table/table.tsx
```diff
@@ -425,9 +433,7 @@ export function Table<T>({
                   })}
                 </tr>
               ))}
-              {selectionEnabled && (
-                <SelectionToolbar table={table} controls={selectionControls} />
-              )}
+              {/* SelectionToolbar row removed from here */}
```
TWilson023: We can remove this comment

packages/ui/src/table/table.tsx
```diff
@@ -325,6 +325,14 @@ export function Table<T>({
             scrollWrapperClassName,
           )}
         >
+          {/* Selection Toolbar Overlay */}
+          {selectionEnabled && (
+            <SelectionToolbar
+              table={table}
+              controls={selectionControls}
+              className="top-0 left-0 right-0"
```
TWilson023: I think these classes are redundant since `SelectionToolbar` already has them on its root element

packages/ui/src/table/selection-toolbar.tsx
marcusljf: Would you suggest just using `h-10` instead, or creating additional variables?

packages/ui/src/table/selection-toolbar.tsx
marcusljf: Just removed it.

packages/ui/src/table/table.tsx
marcusljf: Removed the className.

packages/ui/src/table/selection-toolbar.tsx
marcusljf: Replaced with TW transitions.

packages/ui/src/table/selection-toolbar.tsx
marcusljf: I've set it to `h-11`, but only on the one class. For some reason had it on multiple.

TWilson023 (approved):

PR #2634, opened 2025-07-09
Refactor to `customerExternalId` to be more verbose

steven-tey:


steven-tey:
@coderabbit final review

PR #2654, opened 2025-07-18
Link hover state change

marcusljf:
Right now we're hiding the link selection feature, as it only shows when the user hovers over the circle favicon area.

I've changed it so that the hover state shows when the user hovers on the entire card to give more visibility to the feature.

**Updated**
CleanShot 2025-07-17 at 21 45 07

**Current**
CleanShot 2025-07-17 at 21 45 25


steven-tey:
@coderabbitai full review

steven-tey:
@coderabbitai are there any changes to the actual styling of the card? or just the addition of `CardList.Card.Context.Consumer`?

Also how will this change affect perf?

PR #2682, opened 2025-07-24
Fix discount maxDuration

devkiran:
## Summary by CodeRabbit

* **New Features**
  * Improved handling and normalization of the "amount" field in forms, especially for flat reward or discount types.
  * Default discount type is now set to "percentage" when adding or editing discounts.

* **Bug Fixes**
  * Fixed inconsistent default durations for recurring discounts, now defaulting to a lifetime duration when not specified.


steven-tey:
@coderabbit final review

PR #2685, opened 2025-07-25
improve track open to receive deepLink instead of domain+key

steven-tey:


steven-tey:
@coderabbit final review

any thoughts on the deepLink prop? can you think of a better name?

PR #2764, opened 2025-08-22
Fix a bunch of folders bugs

steven-tey:


steven-tey:
@coderabbitai full review

PR #2818, opened 2025-09-09
Allow automatic lead event creation in track/sale

devkiran:


steven-tey:
@coderabbitai full review

PR #2863, opened 2025-09-19
Improve support for custom commissions in payouts cron

steven-tey:
## Summary by CodeRabbit

- **Bug Fixes**
  - Custom commissions are now always considered for pending and scheduled payouts during a holding period, preventing unintended delays.
  - Standard commissions continue to respect the configured holding period; if the holding period is zero, date-based filtering is not applied.
  - Users may see updated pending amounts and earlier inclusion of custom commissions.
  - No changes to clawbacks, payout creation, or aggregation.


steven-tey:
@coderabbitai full review

PR #2929, opened 2025-10-04
fix(resend): make reply_to snake_case

briankerr:
## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected email payload to use the proper reply-to format for improved compatibility with the email provider.
  - Ensures replies route correctly while retaining the default support address fallback.
  - No user action required; existing behavior remains unchanged.


steven-tey:
Thanks @briankerr! Closing this in favor of #2930 🙏

PR #2937, opened 2025-10-06
Improve discount validation

steven-tey:


steven-tey:
@coderabbitai full review

PR #2943, opened 2025-10-07
Use identityHash for recordClickCache

steven-tey:


steven-tey:
@coderabbitai full review

PR #3233, opened 2025-12-16
chore: upgrade react-email deps

lucasfcosta:
This PR updates the `react-email`-related packages that are likely to cause slow email renders. Those slow renders are probably causing high latency on the routes which call the batch API.

As an evidence, when I run render for 100 emails with your current version of `@react-email/components` and `react-email`, it takes 12s on my Mac. Updating their versions (newer ones shown below) and running the same code takes ~200ms (60x faster).

```
"@react-email/components": "^1.0.1",
"react-email": "^5.0.8",
```

I also did everything I could to verify these on Resend's back-end and that evidence also seems to point this way. That's because the requests we get take ~900ms on the API and your requests only hits us close to the top of the logged timestamps (not when the call to Resend's SDK starts). Therefore, it's more likely that the request is fast but something _before_ it takes long.

## What you should check

Even though I did these updates according to our list of breaking changes, I would still like you to please test your application and have a look at the email templates to make sure they all look okay. I'm not very familiar with your application, so I couldn't do more thorough verification.

Checking the overall email-sending behavior and the email layouts should be enough.

## How to benchmark rendering locally

I benchmarked rendering locally using the script below. In case you want to try it out, you can copy it to the `email` package and run it with `tsx`:

```
import * as React from "react";
import { render } from "@react-email/render";
import PartnerPayoutProcessed from "./src/templates/partner-payout-processed";

async function test() {
  console.log(
    "Rendering 100 emails with actual partner-payout-processed template...",
  );
  const start = Date.now();

  await Promise.all(
    Array.from({ length: 100 }, (_, i) =>
      render(
        <PartnerPayoutProcessed
          email={`test${i}@example.com`}
          program={{
            name: "Test Program",
            logo: "https://assets.dub.co/logo.png",
          }}
          payout={{
            id: `po_test_${i}`,
            amount: 490,
            periodStart: new Date("2024-11-01"),
            periodEnd: new Date("2024-11-30"),
          }}
          variant="stripe"
        />,
      ),
    ),
  );

  console.log(`Done in ${Date.now() - start}ms`);
}

test();
```

## Release Notes

* **Chores**
  * Upgraded email service package dependencies to latest versions for improved stability and performance.

* **Refactor**
  * Enhanced email rendering to properly handle asynchronous operations, ensuring reliable message delivery.


steven-tey:
THank you so much for the PR @lucasfcosta! Closing in favor of #3260 (this PR's build was failing)

Issue #3310, opened 2026-01-03
Account creation shows confusing “Invalid image format” error after Google signup


Puxjee:
### Bug Description
When signing up with Googlethe UI shows an error:
`invalid_string: logo: Invalid image format, supports only png, jpeg, jpg, gif, webp.`

This is confusing because the user never uploads an image.

### Steps to Reproduce
1. Sign up using Google OAuth
2. On the "Create your workspace" screen, fill workspace name + slug (default values)
3. Click "Create workspace"
4. Observe error toast: `invalid_string: logo: Invalid image format...`

### Expected Behavior
Workspace should be created successfully, OR
if logo from Google profile is unsupported, the app should fallback to default avatar/logo without failing workspace creation.

### Actual Behavior
Workspace creation fails and shows a developer-facing validation error.

### Screenshots

<img width="1916" height="903" alt="Image" src="https://github.com/user-attachments/assets/c6751ee5-91d9-46ae-9eba-bf689cc5460b" />

Puxjee:
i’d like to take this. I can submit a PR

devkiran:
@Puxjee We've fixed this issue, thank you for reporting this.

PR #3317, opened 2026-01-06
Fix: Workspace creation fails for Google OAuth users with logo validation error  #3310

mrayushmehrotra:
## Summary by CodeRabbit

* **Bug Fixes**
  * Removed automatic logo population in the workspace creation form from incompatible image sources; logo now requires explicit upload or favicon input.

* **Chores**
  * Updated the zod dependency to a newer patch version to keep libraries current.

### Problem
When users sign up via Google OAuth and attempt to create a workspace without uploading a custom logo, they encounter the error:

> `invalid_string: logo: Invalid image format, supports only png, jpeg, jpg, gif, webp.`

This is confusing because the user never explicitly uploads an image.

### Root Cause
In [create-workspace-form.tsx](cci:7://file:///home/ayush/open/dub/apps/web/ui/workspaces/create-workspace-form.tsx:0:0-0:0), the `useEffect` was setting `logo` to `session.user.image` (the Google profile picture URL, e.g., `https://lh3.googleusercontent.com/...`) for users with generic email domains like `@gmail.com`.

However, the `uploadedImageSchema` only accepts:
- Base64 encoded images
- R2 storage URLs
- Google Favicon URLs (not profile pictures)

The Google profile picture URL doesn't match any of these patterns, causing validation to fail.

### Solution
Removed the fallback that sets the logo to `session.user.image`. Now, users with generic emails will have an empty/undefined logo field, which is valid since the schema uses `.nullish()`. The workspace is created successfully without a logo.

### Changes
- [apps/web/ui/workspaces/create-workspace-form.tsx](cci:7://file:///home/ayush/open/dub/apps/web/ui/workspaces/create-workspace-form.tsx:0:0-0:0) - Removed the `else if (session?.user?.image)` branch that was auto-setting the logo

---

@steven-tey Would appreciate your review on this fix! Let me know if you'd prefer a different approach, such as converting the Google profile image to base64 before submission instead.

image

### DIFF
--- a/apps/web/ui/workspaces/create-workspace-form.tsx
+++ b/apps/web/ui/workspaces/create-workspace-form.tsx
@@ -62,6 +62,8 @@ export function CreateWorkspaceForm({
           // Don't set the logo if fetch fails
           console.log("Failed to check favicon:", error);
         });
+      // } else if (session?.user?.image) {
+      //   setValue("logo", session.user.image);
     }
     // Note: We intentionally don't set the logo to session.user.image (e.g., Google profile picture)
     // because those URLs are not accepted by the uploadedImageSchema which only allows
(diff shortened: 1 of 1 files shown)


mrayushmehrotra:
@coderabbitai do a final review and list out breaking changes, if any

mrayushmehrotra:
@coderabbitai do a final review and list out breaking changes, if any, i've removed the pnpm-lock.yaml file because it can cause error in some case

PR #3350, opened 2026-01-13
Show full ring when the leading side is 100%

marcusljf:
When the program activity is 100% leading in one side or another, so trustworthy or removed instead of showing a little tiny green notch or red notch but it's actually 0, just fill the ring 100%.

### Current
CleanShot 2026-01-13 at 09 13 50@2x

### Updated
CleanShot 2026-01-13 at 09 15 20@2x

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved activity ring rendering logic to better handle edge cases, ensuring visibility of completion indicators in extreme scenarios and maintaining consistent visual representation across different data states.

### DIFF
--- a/packages/ui/src/activity-ring.tsx
+++ b/packages/ui/src/activity-ring.tsx
@@ -105,55 +105,79 @@ export function ActivityRing({
   }, [positiveValue, negativeValue]);
 
   // Calculate arc angles based on ratio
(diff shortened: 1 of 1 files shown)


steven-tey:
Great catch, thank you @marcusljf!

PR #3433, opened 2026-02-06
Updated the character counter direction

marcusljf:
For some reason it was counting down, but it should be counting up to the limit.

| Current | Updated |
| -- | -- |
| CleanShot 2026-02-06 at 15 32 29@2x | CleanShot 2026-02-06 at 15 30 40@2x |

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the description length indicator in the bounty claim modal to display the current character count instead of remaining characters.


steven-tey (approved):

steven-tey:
looks like it's already fixed!

PR #3439, opened 2026-02-08
Fix: build fail with some type error

bikalsiwakoti:
Fixes: #2546

Description:
Fixes some type errors while building file

Make sure it doesn't run out of memory while building.

## Summary by CodeRabbit

* **Refactor**
  * Updated type assertions in the rich text editor components for improved type compatibility. No changes to user-facing functionality or behavior.

### DIFF
--- a/packages/ui/src/rich-text-area/rich-text-provider.tsx
+++ b/packages/ui/src/rich-text-area/rich-text-provider.tsx
@@ -209,7 +209,7 @@ export const RichTextProvider = forwardRef<
               }),
             ]
           : []),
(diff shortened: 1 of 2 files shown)


devkiran:
Thanks for the PR. That said, I'm a bit concerned about adding multiple any casts just to make the build pass. I'd prefer we identify the root cause rather than loosening type safety.

bikalsiwakoti:
Thanks for the feedback!
That makes sense, I agree that using multiple any casts isn’t ideal.

PR #3445, opened 2026-02-09
fix: open Stripe Express payout management in new tab

### LINKED ISSUE #3290: [Feature Request] Manage payout account open in a new tab
In partners.dub.co -> Payouts tab -> Payout Settings -> Connected Payout Account -> Manage, would it be better if it opens Stripe Express on a new tab?

https://github.com/dubinc/dub/blob/db484e7ec450eaf45f4b73d38fa4bac65d93e1d8/apps/web/ui/partners/payout-methods-dropdown.tsx#L197

vishalsaw28:
This updates the Stripe payout account “Manage” action to open Stripe Express in a new tab instead of navigating away from the Dub dashboard.

Since Stripe Express is an external dashboard, opening it in a new tab preserves user context and improves the payout management experience.

Closes #3290

## Summary by CodeRabbit

* **Bug Fixes**
  * Stripe payout method page now opens in a new browser tab instead of navigating within the app, preserving your current session and navigation context.

### DIFF
--- a/apps/web/ui/partners/payout-methods-dropdown.tsx
+++ b/apps/web/ui/partners/payout-methods-dropdown.tsx
@@ -46,7 +46,7 @@ export function PayoutMethodsDropdown() {
           toast.error("Unable to create account link. Please contact support.");
           return;
         }
-        router.push(data.url);
+        window.open(data.url, "_blank", "noopener,noreferrer");
       },
       onError: ({ error }) => {
         toast.error(error.serverError);


steven-tey:
@vishalsaw28 thanks for the PR! We actually had this before but it [gets blocked occasionally by Chrome and other browsers](https://www.reddit.com/r/learnjavascript/comments/13wv1vy/my_browser_is_blocking_my_userscripts_windowopen/), which leads to user confusion, so we're sticking with this for now.

vishalsaw28:
> @vishalsaw28 thanks for the PR! We actually had this before but it [gets blocked occasionally by Chrome and other browsers](https://www.reddit.com/r/learnjavascript/comments/13wv1vy/my_browser_is_blocking_my_userscripts_windowopen/), which leads to user confusion, so we're sticking with this for now.

Got it, that makes sense  thanks for the clarification and context! Appreciate it.

PR #3446, opened 2026-02-09
Redirect paywalls to the correct upgrade plan

marcusljf:
This looks to fix the issue in which users looking to upgrade aren't shown the correct plan to upgrade to based on the paywall they're clicking from.

This create a link type of upgrade_plan={planRecommended}&upgrade_source={sourceLocation} and maps it to the correct version.

CleanShot 2026-02-09 at 15 52 23@2x


marcusljf:
Changing plan.

PR #3453, opened 2026-02-11
Provide context to resolve fraud comments

marcusljf:
`Review feasibility and need`

When fraud reports are resolved without banning the partner, and optional notes are provided, they're added as a comment to the partner comments. But without context, it's not clear what the note is pertaining to.

This addition provides context for any fraud notes left, but only for ones added after this feature is added.

Because old comments are stored `text only` it'll be difficult and not accurate to back fill previous resolution comments to show like this. In old bulk resolve flows, one note could represent multiple resolved fraud groups. The same note text can be reused across many resolutions, so text matching is ambiguous. Without unique linkage in historical data, we can’t always know which exact fraud event a comment came from.

CleanShot 2026-02-10 at 16 06 42@2x

## Summary by CodeRabbit

* **New Features**
  * Partner comments now store structured fraud-resolution metadata and display a “Resolved” banner with rule name plus a View button to open resolution details.
* **Bug Fixes**
  * Improved bulk and single-group resolution to reliably report which groups were resolved and handle concurrent updates.
* **Tests**
  * Added unit tests for serialization/parsing of fraud-resolution comment format.
* **Chores**
  * Added a DB field for comment metadata and a backfill script to migrate legacy comments.

### DIFF
--- a/apps/web/lib/api/fraud/resolve-fraud-groups.ts
+++ b/apps/web/lib/api/fraud/resolve-fraud-groups.ts
@@ -31,7 +31,7 @@ export async function resolveFraudGroups({
       };
     }
 
-    const resolvedAt = new Date();
+    const resolvedAtLowerBound = new Date();
 
     const { count } = await tx.fraudEventGroup.updateMany({
       where: {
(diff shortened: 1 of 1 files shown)


marcusljf:
Converting to a draft, and maybe deleting later as it's part of a much larger platform improvement for fraud detection.

PR #3454, opened 2026-02-11
Improve delete-group feature to be more robust

steven-tey:
## Summary by CodeRabbit

* **Refactor**
  * Partner reassignment during group deletion is now batched and processed in background; final audit logging simplified.
* **Bug Fixes / Cleanup**
  * Discount remapping and cleanup moved out of the immediate delete transaction into the background job; unused discounts are removed after reassignment when applicable.
* **Chore**
  * A deletion flag is propagated through partner-move and remap workflows.


steven-tey:
@coderabbitai full review, are there any breaking changes?

PR #3462, opened 2026-02-12
Bulk partner invite

marcusljf:
Allows the workspace to invite one to many partners (max 50) with the same application instance. The user separates each email with a comma to create the email chip.

Additional features:
- Delete to highlight the previous email chip. Press key again to delete
- User can use the left and right arrow keys to select other chips to delete
- When tabbing out of the input, the email chip is created or validated to correct
- Separate emails are sent to each partner to not share PIP
- User can click on the `X` on each email chip to delete from the list as well.

https://github.com/user-attachments/assets/650b9e54-ea9b-40a2-8b7b-0f612dc74a1e

## Summary by CodeRabbit

* **New Features**
  * Multi-email recipient input with chips, CSV paste, keyboard navigation and a 50-recipient limit
  * Bulk partner invites with aggregated success/skip feedback and batched personalized emails

* **Improvements**
  * Show name/short-link fields only for single recipients
  * Better helper text, per-recipient validation, combined Send/Cancel loading and disabled states
  * Email preview preserved and integrated with multi-recipient flow

### DIFF
--- a/apps/web/app/app.dub.co/(dashboard)/[slug]/(ee)/program/partners/invite-partner-sheet.tsx
+++ b/apps/web/app/app.dub.co/(dashboard)/[slug]/(ee)/program/partners/invite-partner-sheet.tsx
@@ -26,7 +26,7 @@ import {
   Sheet,
   useMediaQuery,
 } from "@dub/ui";
-import { cn } from "@dub/utils";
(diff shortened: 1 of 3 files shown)


steven-tey (approved):

steven-tey:

PR #3490, opened 2026-02-20
Sweeping table improvements

marcusljf:
Improvements to the table view including:
- Unified spacing on the checkbox column, so it's the same on all tables. If a column is next to a checkbox, it reduces the left padding but reverts to default when there's no checkbox column.
- Unified width and padding on the `⋯` action column. On some tables, the `⋯` action was it was much wider than it needed to be.
- Fixed alignment on the multi-select header actions. So when you select one of the check boxes or all the checkboxes above, it is in line with everything and the text is in line with the column next to it. 
- Increased the clickable safe area of the checkbox on the tables to reduce misclicks when selecting rows. I often found that if I didn't click exactly in the check box, it would open up that page. If I'm experiencing it, somebody else is going to. 

https://github.com/user-attachments/assets/cbb600d0-aed7-41d1-9a11-ef255c90bf97

This work also helped resolve the issue where the fraud table had different spacing compared to every other table so all the tables look alike now.

CleanShot 2026-02-20 at 15 21 54@2x

## Summary by CodeRabbit

* **Style**
  * Refined button appearance across dashboard menus for more consistent sizing, spacing and rounded corners.
  * Adjusted partner row layout and tooltip behavior for clearer avatar and status display.
  * Increased location column width in partner tables for better content display.

* **New Features**
  * Enhanced row selection UI with a consolidated select-all control, shift-click range selection and animated selection label.
  * Improved table layout with fixed utility column sizing, responsive column scaling, and more stable rendering.


steven-tey:
@coderabbitai full review, any breaking changes (both UI and functionality wise)?

steven-tey (approved):

PR #3567, opened 2026-03-11
Guard in recordClickCache

steven-tey:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache error handling so cache access failures are swallowed and logged, preventing request interruptions.
  * Ensures cached click detection short-circuits processing on hits to avoid duplicate work.
  * Prevents downstream overload when cache is unavailable by returning early and preserving normal request flow.

* **Improvements**
  * Faster local cache updates and consistent invalidation of edge cache after link updates.
  * Added an edge-fallback to fetch and cache missing links on cache miss.


steven-tey:
@coderabbitai full review, any edge cases/breaking changes?

PR #3619, opened 2026-03-20
perf: add CDN cache headers to image generation endpoints

JohnPhamous:
## Summary by CodeRabbit

* Improved caching for avatar and QR code endpoints to enhance performance and reduce response times on repeated requests.

---

## Summary
- Add `Cache-Control` and `Vercel-CDN-Cache-Control` headers to `/api/og/avatar` and `/api/qr`
- Both generate deterministic images (same input = same output) but currently serve uncached

## Changes
- **`/api/og/avatar`**: Immutable cache (1yr). Deterministic based on seed.
- **`/api/qr`**: 24h CDN + 1h browser cache. Shorter TTL since QR logo can change on plan upgrade.

## Impact
Avoids redundant `ImageResponse` (Satori) rendering on repeated requests.

### DIFF
--- a/apps/web/app/api/qr/route.tsx
+++ b/apps/web/app/api/qr/route.tsx
@@ -26,13 +26,6 @@ export async function GET(req: NextRequest) {
 
     const qrCodeLogo = await getQRCodeLogo({ url, logo, hideLogo });
 
-    const responseHeaders = new Headers(CORS_HEADERS);
-    responseHeaders.set(
-      "Vercel-CDN-Cache-Control",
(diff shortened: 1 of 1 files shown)


JohnPhamous:
Re: header construction pattern nitpick — the difference is because `CORS_HEADERS` in the QR route is a `Headers` object (needed for the OPTIONS handler) while the avatar route uses a plain object spread. Keeping as-is since changing would also need to update the OPTIONS handler and error handler that share `CORS_HEADERS`.

PR #3660, opened 2026-03-27
Fix update domain for partnerProgram

steven-tey:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed domain rename flow to ensure cloud resources are removed, renames are queued for follow-up, and partner-program records are updated when a domain is changed.
  * Improved cache invalidation so only affected domains and well-known paths are revalidated on updates.
  * Corrected domain renewal calls to use the correct/current domain value.
  * Cron verification now excludes domains associated with partner programs to prevent unintended status updates and reminders.


steven-tey:

PR #3675, opened 2026-04-01
Fix bounty maxSubmissions test to use correct max value

devkiran:
## Summary by CodeRabbit

* **Tests**
  * Updated validation tests for the bounties endpoint's maximum submissions limit.


steven-tey:
Fixing in #3679

PR #3733, opened 2026-04-10
Block saving additional links when enabled without a link format

pepeladeira:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented saving group links when additional link formats are enabled but none are configured by skipping the update.
  * Disabled the Save button (with tooltip) when a link format is required but missing.
  * Reordered link settings so the "Link limit" appears after "Link formats" for clearer workflow.


steven-tey:
@coderabbitai full review

steven-tey (approved):

PR #3863, opened 2026-05-06
Fix workspace delete cron stopping after the first batch

devkiran:
The bug is the cursor pagination. Look at the sequence:

Batch 1: fetches links 1..100 (sorted by id asc), deletes them all, sets startingAfter = id_of_link_100.

Batch 2 runs with cursor: { id: id_of_link_100 }, skip: 1 — but id_of_link_100 no longer exists because we just deleted it.

## Summary by CodeRabbit

* **Refactor**
  * Simplified workspace deletion processes by removing cursor-based pagination logic from customer, domain, folder, and link deletion operations. The system now processes workspace cleanup in consistent batches without pagination markers while maintaining all deletion functionality.


devkiran:
@coderabbitai full review please

steven-tey (approved):

PR #3883, opened 2026-05-11
fix: return 200 on business-logic errors in Stripe webhook handler

### LINKED ISSUE #3752: Stripe webhook handler returns 4xx on business logic errors, causing unguarded retries
## Bug Description

The Stripe webhook handler at `apps/web/app/(ee)/api/stripe/webhook/route.ts` returns HTTP 400 on business logic errors (e.g., email send failure). Stripe treats 4xx as "retry me," which re-fires events like `checkout.session.completed` with no idempotency guard — potentially causing double plan upgrades or duplicate processing.

**Affected file:** `(ee)/api/stripe/webhook/route.ts` (around line 83)

## Expected Behavior

Stripe webhook handlers should return HTTP 200 for all successfully *received* events, regardless of whether downstream business logic succeeds. Business logic failures should be handled internally (retry queue, dead-letter, logging) — not surfaced as HTTP status codes that trigger Stripe's retry machinery.

## Suggested Fix

1. Return 200 for all received events:
```ts
try {
  await handleEvent(event);
} catch (error) {
  // Log error, enqueue for retry internally — but don't tell Stripe to retry
  console.error("Webhook processing failed:", error);
  await deadLetterQueue.add(event);
}
return NextResponse.json({ received: true }, { status: 200 });
```

2. Add event-ID deduplication (prevents double-processing on legitimate Stripe retries):
```ts
const eventId = event.id;
const alreadyProcessed = await redis.set(`stripe:event:${eventId}`, "1", { nx: true, ex: 86400 });
if (!alreadyProcessed) {
  return NextResponse.json({ received: true, deduplicated: true }, { status: 200 });
}
```

## Impact

Without this fix, any transient downstream failure (email service timeout, database blip) causes Stripe to re-deliver the event up to its retry limit, and each delivery re-executes the full handler including subscription updates and workspace plan changes.

---
*Found via [ADQB](https://github.com/Nebutra/adqb) architecture audit*

brone1323:
## What

Changes the business-logic catch block in `apps/web/app/(ee)/api/stripe/webhook/route.ts` to return `status: 200` instead of `status: 400`.

## Why

Per [Stripe's webhook docs](https://stripe.com/docs/webhooks#acknowledge-events-immediately), any non-2xx response signals that the event was not received and triggers automatic retries. When a handler throws due to a transient error (e.g. an email delivery failure), the current 400 response causes Stripe to re-deliver the event repeatedly — potentially resulting in duplicate subscription upgrades, charges, or other idempotent-sensitive operations.

Signature verification errors correctly keep `400` (those are genuine bad requests). Internal handler errors should acknowledge receipt and rely on our own alerting/logging for investigation.

Closes #3752

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated webhook handling to return correct HTTP status codes for processing failures, preventing redundant retries and improving webhook reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dubinc/dub/pull/3883)

### DIFF
--- a/apps/web/app/(ee)/api/stripe/webhook/route.ts
+++ b/apps/web/app/(ee)/api/stripe/webhook/route.ts
@@ -83,8 +83,10 @@ export const POST = async (req: Request) => {
       message: `Stripe webhook failed (${event.type}). Error: ${error.message}`,
       type: "errors",
     });
+    // Return 200 so Stripe does not retry on business-logic failures.
+    // Signature/construction errors (above) still return 400, which is correct.
     return new Response(`Webhook error: ${error.message}`, {
-      status: 400,
+      status: 200,
     });
   }
 


devkiran:
Duplicate https://github.com/dubinc/dub/pull/3773

PR #3947, opened 2026-05-25
Program Marketplace Categories

pepeladeira:
## Summary by CodeRabbit

* **New Features**
  * Marketplace relocated to `/marketplace` with enhanced filtering, sorting, and category browsing.
  * New program discovery experience including featured programs carousel, category navigation, and refined list views.
  * Program detail pages with improved hero sections, eligibility information, and call-to-action placement.

* **Refactor**
  * Middleware routing updated to support new marketplace URL structure and legacy URL redirects.
  * Sidebar navigation restructured with dedicated marketplace panel and improved layout controls.


apps/web/ui/program-marketplace/featured-program-card.tsx
```diff
@@ -0,0 +1,234 @@
...
+                    ?.map((category) => (
+                      <ProgramCategory
+                        key={category}
+                        category={category}
+                        variant="surface"
+                        onClick={() =>
+                          router.push(getMarketplaceCategoryHref(category))
+                        }
+                      />
+                    ))}
```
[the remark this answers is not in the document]
pepeladeira: It's already inside the ProgramCategory component.

apps/web/ui/program-marketplace/featured-program-card.tsx
```diff
@@ -0,0 +1,234 @@
...
+                            <ProgramCategory
+                              key={category}
+                              category={category}
+                              variant="surface"
+                              onClick={() =>
+                                router.push(
+                                  getMarketplaceCategoryHref(category),
+                                )
+                              }
+                            />
```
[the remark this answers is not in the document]
pepeladeira: It's already inside the ProgramCategory component.

pepeladeira:
@coderabbitai full review. any breaking changes or security issues?

apps/web/lib/middleware/utils/partners-redirect.ts
```diff
@@ -14,6 +15,62 @@ export const partnersRedirect = (path: string) => {
...
+    return withQuery("/marketplace", searchParams);
+  }
+
+  if (path === "/programs/marketplace/all") {
+    return withQuery("/marketplace/all", searchParams);
+  }
+
+  if (
+    path === "/programs/marketplace/popular" ||
+    path === "/marketplace/popular"
```
general-adhoc, line 50: Is `/marketplace/popular` is intended to be a standalone “Popular programs” page?

Steven’s URL design described this as `Popular programs (take top 12 programs by ranking) -> /marketplace/popular`, but the current implementation redirects it to `/marketplace/all?sortBy=popularity&sortOrder=desc`. 

The destination is the generic page w/ "All Programs" / "Find your next partnership" copy and canonicalizes as `marketplace/all` so our new `popular` page won't behave like a distinct popular SEO page / Google won't register it in our sitemap 

If the product decision is that `/popular` is only an alias for the sorted all-programs list, then this is fine, but we may not see any SEO benefits from this one.  Would recommend we get this registering as a standalone page for sitemap benefit

apps/web/ui/program-marketplace/external/marketplace-external-shell.tsx
```diff
@@ -0,0 +1,59 @@
...
+        {children}
+      </div>
+    );
+  }
+
+  const resolvedTitle =
+    title ??
+    (variant === "home" ? (
+      <>
+        Dub Program
```
general-adhoc: nit: Is "Program Marketplace" the top keyword we are attempting to rank for?  

Would recommend something more contextually relevant for H1 like "Affiliate Program Marketplace" or "Influencer Program Marketplace".

apps/web/app/app.dub.co/marketplace/[[...slug]]/page.tsx
```diff
@@ -0,0 +1,93 @@
...
+    canonicalUrl: getMarketplaceCanonicalUrl(pathname),
+  });
+}
+
+export default async function MarketplaceExternalPage(props: {
+  params: Promise<{ slug?: string[] }>;
+  searchParams: Promise<Record<string, string | string[] | undefined>>;
+}) {
+  const { slug } = await props.params;
+  const searchParams = await props.searchParams;
```
Comment (general-adhoc) [inline apps/web/app/app.dub.co/marketplace/[[...slug]]/page.tsx]: This unconditional `await props.searchParams` appears to opt the marketplace pages out of ISR. 

I tested this against a production build locally: `/marketplace` rewrites to `/app.dub.co/marketplace`, but the response header is `Cache-Control: private, no-cache, no-store, max-age=0, must-revalidate`, and the marketplace routes are missing from `.next/prerender-manifest.json` despite `revalidate = 3600`.

Image

That means these SEO pages are rendered dynamically on every request instead of being served from the ISR cache, so crawler traffic can hit the DB/render path directly. Can we avoid reading `searchParams` at the top level and only read it in the branches that actually need query params, or split the query-param/filter routes from the static marketplace landing/category/program pages?

apps/web/ui/program-marketplace/external/marketplace-external-list-page.tsx
```diff
@@ -0,0 +1,141 @@
...
+export async function MarketplaceExternalListPage({
+  slug,
+  searchParams,
+  fixedCategory,
+}: {
+  slug?: string[];
+  searchParams: Record<string, string | string[] | undefined>;
+  fixedCategory?: Category;
+}) {
+  const { rewardType, search, sortBy, sortOrder, page } =
```
general-adhoc: nit: Invalid query params currently crash the public marketplace page. 

For example, `/marketplace/all?page=0` throws from this `.parse(...)` and renders Next's production “Application error” page; in my local prod build the document still returned `200 OK`, so crawlers could receive a successful status with an error page body.

Since this is parsing user-controlled query params in a public SEO page rather than an API handler, can we use Zod `safeParse` here and redirect to the cleaned/default URL when params are invalid?

Image

apps/web/lib/fetchers/get-public-network-programs.ts
```diff
@@ -0,0 +1,185 @@
...
+          include: {
+            clickReward: true,
+            leadReward: true,
+            saleReward: true,
+            referralReward: true,
+            discount: true,
+          },
+        },
+        categories: true,
+        invoices: true,
```
general-adhoc: v minor scaling/performance nit:

adding `invoices:true` includes every invoice row for each program as a popularity tie-breaker.

Likely not an issue today, but for these public SEO pages, this can get expensive as invoice history grows: we're loading full invoice records/columns just to compute a `SUM(amount)` in JS. 

Simple fix is to replace this with a DB aggregate (SUM) and avoid `invoices: true` on the SSR path

apps/web/app/app.dub.co/marketplace/[[...slug]]/page.tsx
Comment (steven-tey) [reply apps/web/app/app.dub.co/marketplace/[[...slug]]/page.tsx]: Agreed, `searchParams` should be handled client-side – right now, navigating the marketplace feels quite sluggish because of the ISR opt-out 😅 

cc: @pepeladeira

apps/web/lib/middleware/utils/partners-redirect.ts
steven-tey: Great catch, though I think I'm OK with not having a separate `/popular` page since it's just a sorted version of the all-programs list and hence won't have much SEO benefit 👍

apps/web/ui/program-marketplace/external/marketplace-external-list-page.tsx
steven-tey: Great catch, cc: @pepeladeira

apps/web/lib/fetchers/get-public-network-programs.ts
steven-tey: Agreed, let's optimize this 👍

steven-tey: removing sorting by invoices for now since we're mostly relying on `marketplaceRanking` atm 👍

PR #3953, opened 2026-05-27
Extra ssrf guards

pepeladeira:
## Summary by CodeRabbit

* **Bug Fixes**
  * Safer outbound requests: comprehensive SSRF protections, private-IP blocking, stricter redirect and timeout handling, and explicit failures for invalid/non-OK targets.

* **Refactor**
  * Iframe eligibility now evaluates response headers with request origin for more reliable embedding decisions.
  * Multiple API flows and sitemap fetching updated to use the safer outbound fetch behavior.

* **Tests**
  * Added private-IP detection tests and adjusted sitemap/content fetch mocks; extended test logging mocks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dubinc/dub/pull/3953?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


pepeladeira:
@coderabbitai full review. any breaking changes? any security issues?

apps/web/lib/api/safe-fetch.ts
```diff
@@ -0,0 +1,297 @@
...
+    if (remaining <= 0) {
+      throw new Error("Request timed out");
+    }
+
+    const controller = new AbortController();
+    const timer = setTimeout(() => controller.abort(), remaining);
+
+    let response: Response;
+    try {
+      response = await fetch(currentUrl.toString(), {
```
sviat-barbutsa, line 268: Nice direction. I think there is still a TOCTOU gap here: `assertUrlIsSafe(currentUrl)` validates the hostname through DoH, but the actual `fetch(currentUrl.toString(), ...)` lets the runtime resolver resolve the same hostname again.

It means the checked IPs are not pinned to the request. A host that gives a public answer during the DoH check but resolves to an internal address for the actual fetch, or changes between the check and the fetch, can still bypass the SSRF guard.

Could you please either make the request use the already-validated address, while preserving the original `Host`/TLS SNI semantics, or add a test that demonstrates this rebinding/split-DNS case is blocked? The redirect loop already revalidates each hop, so this is the remaining weak spot I see in the helper.

PR #3979, opened 2026-06-04
Fix commission creation workflow

steven-tey:
## Summary by CodeRabbit

* **Updates**
  * Customer name field is now optional, allowing records to be created or updated without a provided name.
  * Discount codes tied to program enrollments are now automatically removed when the associated enrollment is deleted.
* **Bug Fixes**
  * Customer filter labels now fall back to customer ID when name and email are unavailable, avoiding blank labels.


steven-tey:
@coderabbitai full review

PR #3985, opened 2026-06-04
Fix discount code Stripe/DB drift on creation failure

pepeladeira:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate discount-code detection to use the correct program matching.
  * Prevented assigning more than one discount code to the same link.
  * Better error responses for conflicts when external provider or storage report a taken code.
  * Safer creation flow with rollback of external codes if persisting fails, and clearer conflict handling.


pepeladeira:

steven-tey (approved):

PR #3997, opened 2026-06-08
Fix Android Play Store fallback for in-app webviews

pepeladeira:
## Summary by CodeRabbit

* **Bug Fixes**
  * Unified fallback URL for deep links so all navigation branches use the same fallback.
  * On Android, intent paths include the unified fallback and a short timeout that redirects if the intent doesn't take over; the timeout is cleared when the page becomes hidden.
  * Other browsers (including Firefox on Android) now consistently navigate to the same fallback URL.


steven-tey:
@coderabbitai full review, any breaking changes in behavior?

steven-tey (approved):

PR #3999, opened 2026-06-08
Add per-IP rate limit to OTP

pepeladeira:
## Bug Fixes
* OTP requests now use a dual-tier rate-limiting strategy, combining a tighter per-identifier limit with an additional IP-based limit to better curb abuse without disrupting legitimate traffic.  
* Client IP detection has been improved using more reliable request header resolution, with clearer provenance handling and safer fallback behavior when an IP can’t be determined.  
* When OTP limits are exceeded, the user-facing “Too many requests. Please try again later.” message is returned consistently.


apps/web/lib/actions/send-otp.ts
```diff
@@ -30,11 +30,15 @@ export const sendOtpAction = actionClient
   .action(async ({ parsedInput }) => {
     const { email } = parsedInput;
 
-    const { success } = await ratelimit(2, "1 m").limit(
-      `send-otp:${email}:${await getIP()}`,
-    );
+    const ip = await getIP();
```
general-adhoc: `getIP()` can silently return `0.0.0.0` when request IP headers are missing. Since this PR adds an IP-only limiter, fallback usage becomes much more important to observe. Could we expose/log the IP source here, so we can monitor how often OTP requests are falling into the shared fallback bucket?

apps/web/lib/actions/send-otp.ts
```diff
@@ -30,11 +30,15 @@ export const sendOtpAction = actionClient
...
 
-    if (!success) {
+    const [{ success: emailSuccess }, { success: ipSuccess }] =
+      await Promise.all([
+        ratelimit(2, "1 m").limit(`send-otp:${email}:${ip}`),
+        ratelimit(15, "1 h").limit(`send-otp:${ip}`),
+      ]);
+
+    if (!emailSuccess || !ipSuccess) {
       throw new Error("Too many requests. Please try again later.");
```
general-adhoc, line 72: would recommend logging more detailed info when we are rejecting OTP requests

general-adhoc (changes requested):
Overall I think this is a great system protection to add (esp. against bots / AI), however would recommend adding a little better logging/observability so we can understand how often we're using `FALLBACK_IP_ADDRESS` and denying OTP requests.

Also, it would be good to setup some sort of alarming on OTP throttling so we can understand if we are getting attacked vs. many new signups for a new corp client etc.  Maybe >10 blocks in 10m to start

pepeladeira:
@coderabbitai full review

apps/web/lib/actions/send-otp.ts
```diff
@@ -30,11 +32,34 @@ export const sendOtpAction = actionClient
...
+    }
+
+    const [{ success: emailSuccess }, ipResult] = await Promise.all([
+      // Rate limit by email and IP
+      ratelimit(2, "1 m").limit(`send-otp:${email}:${ip}`),
+      // Skip IP-only limit when IP is unknown to avoid a shared global bucket
+      isFallbackIp
+        ? Promise.resolve({ success: true as const })
+        : ratelimit(15, "1 h").limit(`send-otp:${ip}`),
+    ]);
```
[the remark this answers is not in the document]
pepeladeira: On `main` it was already:
```
send-otp:${email}:${await getIP()}
```
This PR kept that key and added an IP-only layer. So this is not a regression introduced by your change.

PR #4002, opened 2026-06-09
Fix datetimeFieldsUpdated logic in updateWorkspacePlan

steven-tey:
## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable workspace billing sync with the payment provider: billing-cycle end dates, cancellation dates, and Stripe-provided trial end dates are now handled correctly during partial updates to prevent stale or incorrect billing or trial states. Logging clarifies when plan/datetime updates are skipped.


steven-tey:

PR #4012, opened 2026-06-10
Skip instead of throwing when transferable amount is below min withdrawal amount

devkiran:
## Summary by CodeRabbit

* **Bug Fixes**
  * Transfers and payouts below the minimum threshold now log a warning and are skipped by default to avoid spurious failures; when explicitly forced, the operation will raise an error instead. This applies to both fiat (Stripe) transfers and stablecoin payouts.


devkiran:
@coderabbitai full review please

steven-tey (approved):

PR #4025, opened 2026-06-11
Message attachment style adjustments

marcusljf:
- Aligned the paper clip button to the rest of the button, and matched the size and styling as the others
- Set it so tailwind can pickup the other library colors that were being used in the designs (rose-600)
- PDF block
  - Moved the delete to the top right to match the image delete
  - Reduced the padding around the icon area
  - The sent document version match the design closer now
  - Also has the correct height. If the pdf was the only attachment it was a smaller height. When an image was added as well, it was taller.

image

## Summary by CodeRabbit

* **Style**
  * Updated PDF attachment color for improved visual consistency.
  * Refined spacing, padding, and layout of attachment buttons, file-type badges, and attachment chips.
  * Switched attachment toolbar control to a standardized attach-file button.

* **Bug Fixes / Behavior**
  * Partner-count grouping no longer provides platform or subscriber-group facets; only status and country groupings remain.


steven-tey (approved):

steven-tey:
Love the @t3dotgg cameo 😍

PR #4032, opened 2026-06-11
Lead and sales rewards metadata condition

pepeladeira:
## Release Notes

* **New Features**
  * Added metadata-based reward conditions for **lead** and **sale**, including metadata field input and new operator options like `contains` / `not_contains`.

* **Bug Fixes**
  * Improved reward-condition evaluation with stricter handling for empty/invalid values and better type-appropriate comparisons for metadata.
  * Updated partner commission queuing to include metadata only when present; tightened sale/lead metadata aggregation.

* **Tests**
  * Expanded unit tests for metadata scenarios, operator edge cases, and schema validation.

* **Chores**
  * Added `QSTASH_URL` to E2E workflows and the web environment example.


devkiran (approved):

apps/web/lib/api/rewards/reward-condition-metadata.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  if (typeof fieldKey === "string") {
+    if (fieldKey.trim() === "" || Number.isNaN(Number(fieldKey))) {
+      return undefined;
+    }
+
+    return Number(fieldKey);
+  }
+
+  const n = Number(fieldKey);
```
sviat-barbutsa, line 32: Nice addition but I think the numeric metadata path needs one more edge-case test for non-scalar JSON values.

Simply because `trackLead` / `trackSale` accept metadata as `z.record(z.string(), z.any())`, so arrays can reach this helper. Because this fallback uses `Number(fieldKey)`, values like `[]` become `0` and `["10"]` becomes `10`, meaning a metadata condition such as `sale.metadata.seats > 5` could match an array value by accident.

Could you, please, guard numeric metadata coercion to strings/numbers only, and add a regression test that array metadata values do not satisfy numeric operators?

devkiran: Thanks for the suggestion @sviat-barbutsa

steven-tey (approved):

PR #4049, opened 2026-06-17
dedup @prisma/client to fix intermittent PlanetScale `:vN` query errors

general-adhoc:
intermittent PlanetScale syntax errors (e.g., syntax error near ':v7') query bug has returned.  Found additional issues with our Prisma setup.

Changes below:

## 1. Externalize Prisma from the Next server bundle

Next already treats `@prisma/client` and `prisma` as server-external packages by default, per the [serverExternalPackages docs](https://nextjs.org/docs/app/api-reference/config/next-config-js/serverExternalPackages). This PR applies the same treatment to our `@dub/prisma` wrapper package so the generated Prisma client stays on the native Node resolution path instead of being bundled into Next server chunks.

## 2. Split Prisma Client re-exports into type-only exports and runtime value exports.

`packages/prisma/client.ts` previously used `export * from "@prisma/client"`, which re-exported the full generated Prisma Client surface at runtime. That includes model types, Prisma namespaces, generated enums, and client values. In the Next/Turbopack dev runtime, that broad re-export over Prisma's generated client can produce unstable runtime behavior.

## 3. packages/prisma was pinned to typescript@5.2.2 while the rest of the workspace used ^5.4.4+. 

Because typescript is a peer dependency of @prisma/client, pnpm installed two separate instances of the Prisma client.  Unclear how large of an impact this had as these clients shouldn't be used at the same time (next-auth vs dub), but cleaning up as part of this PR

## Dev Instructions

First-time sequence after pulling/applying (from workspace root)
```
  pnpm install                         
  pnpm --filter=@dub/prisma generate
  rm -rf apps/web/.next                 # drop stale bundled-Prisma chunks
  pnpm dev                              
```

## Summary by CodeRabbit

* **Chores**
  * Updated Prisma dependencies to the latest version.
  * Optimized dependency bundling configuration for improved server-side performance.
  * Updated TypeScript to version 5.4.4.


steven-tey:
Thank you! Closing this as we move to #4055 instead 🙏

PR #4072, opened 2026-06-24
Search bar performance improvements

general-adhoc:
### Summary
I noticed the search bars felt laggy across multiple screens (originally partner network search).  Root caused it to the search bar component which impacted all 20 usages of `search-box.tsx` (19 of `SearchBoxPersisted` one of `SearchBox` directly)

Separating this out from the partner network search PR as this has a larger potential blast radius than just partner search.

**Two Problems**
1. Pressing enter on any search bar after entering a search term still waited out the 500ms default debounce duration (i.e., enter key did not flush debounced content)
2. committing a search triggers a **server round-trip** (RSC re-render) that produced no data because the current `SearchBoxPersisted` views render results from **client** SWR fetches.

Note: several components were worse with additional non-shallow write (tags, folders, domains, and admin network)

This caused an up to 500ms debounce delay + server round trip for RSC re-render BEFORE the client would begin fetching data, leading to noticeable lag.  

```
BEFORE:
type search term + enter -> wait 500 ms -> router.push -> RSC round-trip (wasted) -> SWR fetch -> render

AFTER:
type search term + enter -> history.pushState -> SWR fetch -> render

```

### Reproduce
1) go to any search bar (try links - https://app.dub.co/{{your-workspace}}/links, quickly type something and press enter)
2) you'll notice the debounce + RSC re-render before the app starts your search (you can also see the _rsc network request - see below for example screenshot)

### Solution
- flush debounce on user enter
- added a new opt-in `shallow?: boolean` on the shared `queryParams` hook - this updates URL via history API instead of router.  This avoids triggering navigation, RSC round trips.  'useSearchParams' still reacts and SWR refetches.
- `SearchBoxPersisted` opts in to shallow by default.  This is safe across all current usages.  
- Cleaned up redundant `onChangeDebounced` handlers
- added regression test ensuring `shallow` write uses history api and not router
- **BEHAVIOR CHANGE**: `queryParams` skips when resulting URL is unchanged - prevent wasted same-URL navigation.  This should be safe, want to call out.

image

## Summary by CodeRabbit

* **New Features**
  * Search filtering can now use shallow URL updates to feel more immediate and avoid full navigation.
  * Search inputs now more reliably apply typed queries when pressing Enter.

* **Bug Fixes**
  * Persisted search now keeps the displayed value and URL in sync without interrupting active typing, and clears pagination when updating the query.
  * Improved handling of clearing/changing search to update results more consistently.

* **Tests**
  * Added coverage for shallow routing behavior in URL updates.

### DIFF
--- a/apps/web/app/(ee)/admin.dub.co/(dashboard)/partners/network/page.tsx
+++ b/apps/web/app/(ee)/admin.dub.co/(dashboard)/partners/network/page.tsx
@@ -139,6 +139,7 @@ export default function NetworkApplicationsPage() {
     queryParams({
       del: ["networkStatus", "country", "page"],
       scroll: false,
+      shallow: true,
     });
   };
 
--- a/apps/web/app/app.dub.co/(dashboard)/[slug]/(ee)/program/campaigns/[campaignId]/campaign-events-modal.tsx
+++ b/apps/web/app/app.dub.co/(dashboard)/[slug]/(ee)/program/campaigns/[campaignId]/campaign-events-modal.tsx
@@ -7,7 +7,6 @@ import { buildUrl, fetcher } from "@dub/utils";
 import { useParams, useRouter } from "next/navigation";
 import { useState } from "react";
 import useSWR from "swr";
-import { useDebouncedCallback } from "use-debounce";
 import { CampaignEvent, EventStatus } from "./campaign-events";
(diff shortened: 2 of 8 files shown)


general-adhoc:
@coderabbitai full review

any regressions?

PR #4095, opened 2026-06-29
Fix duplicate partners in paginated list and API

pepeladeira:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `/partners` offset pagination to return results in a deterministic, stable order across pages by adding an `id` tie-breaker when multiple records share the same sort value.
* **Tests**
  * Added concurrent end-to-end integration tests for `/partners` pagination that request multiple offset pages, verify item counts, ensure no overlap between pages, and confirm the combined ordering matches the expected baseline sequence.


pepeladeira:

PR #4136, opened 2026-07-06
Cancel trialing subscriptions when trial payment validation fails

pepeladeira:
## Summary by CodeRabbit

* **New Features**
  * Added handling for Stripe trial-ending events to automatically review payment readiness.
  * Trial subscriptions now explicitly cancel if no valid payment method is available.

* **Bug Fixes**
  * Improved trial checkout behavior by validating payment methods before attempting to continue billing.
  * Added safer handling for unexpected Stripe payment outcomes during trial conversion.


steven-tey:
@pepeladeira just spoke with Stripe and they're planning to improve billing analytics, so we might not actually need this PR 👀

PR #4188, opened 2026-07-21
Migrate link tag deletion to new background job

devkiran:
## Summary by CodeRabbit

* **New Features**
  * Tag deletion now triggers an asynchronous background job to remove tag-link associations and finalize tag cleanup.
  * Large numbers of associations are processed in batches, reducing impact on interactive requests.
  * When the tag is eligible, affected links are updated as associations are removed, and the tag is deleted.

* **Bug Fixes**
  * Deletion is more robust when the tag is missing, already ineligible, or when only part of the association set remains.
  * Existing in-flight cron messages continue to be handled during the migration window.


steven-tey (approved):

steven-tey:
@coderabbitai full review

steven-tey (approved):

Issue #4200, opened 2026-07-22
Prevent rendering 0 from falsy totalLinks check


Prasad48432:
## Summary

As a keen observer on a random Saturday, I spotted an unexpected `0` sneaking into the UI 😂.

<img width="1080" height="2096" alt="Image" src="https://github.com/user-attachments/assets/50245221-b5bf-4675-b995-1260d65f03ee" />
 

The culprit was the conditional rendering:

## Links

- **File:** `apps/web/ui/links/link-display.tsx`
- **Line:**  83

```tsx
{totalLinks && totalLinks < SORTABLE_LINKS_LIMIT && (...)}
```

Fix:
```tsx
{totalLinks > 0 && totalLinks < SORTABLE_LINKS_LIMIT && (...)}
```

pepeladeira:
Hi @Prasad48432! Thank you for flagging it.

We're fixing it here: https://github.com/dubinc/dub/pull/4201

gonzoblasco:
Fix is up on `gonzoblasco:fix/totalLinks-zero-4200`. Simple one-liner: changed `totalLinks &&` to `totalLinks > 0 &&` so it does not render `0` when `totalLinks` is falsy.

steven-tey:
Thank you for reporting this, just pushed up a fix!

PR #4201, opened 2026-07-22
Prevent rendering 0 from falsy totalLinks check

pepeladeira:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the visibility of link sorting controls.
  * Sorting options are now shown only when links are available and the total count is within the supported limit.


steven-tey (approved):

steven-tey:
closes #4200

PR #4239, opened 2026-07-29
Fix URI malformed errors when decoding link keys

devkiran:
Before: malformed % in a key → uncaught URIError → 500

After: same key → decode skipped → lookup as-is → usually not_found (or whatever the existing miss path is)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed URL-encoded domains and keys without causing lookup errors.
  * Link lookups now remain functional when percent-encoded values are invalid or incomplete.

* **New Features**
  * Added safer URL decoding that preserves the original value when decoding fails.

* **Tests**
  * Added coverage for valid, malformed, plain-text, and non-ASCII URL values.


steven-tey (approved):

steven-tey:
@coderabbitai full review

PR #4248, opened 2026-07-30
Make merge accounts more scalable for `commissions` and `customers`

steven-tey:
Also added `PRISMA_UPDATEMANY_LIMIT` const

## Summary by CodeRabbit

* **Improvements**
  * Standardized batch processing limits across scheduled cleanup jobs, fraud commission holds/releases, migration backfills, and partner/program data maintenance tasks.
  * Improved consistency when processing large datasets by using a shared batching limit instead of hardcoded values.
  * Updated the account-merge batching behavior to process large updates in controlled chunks with clearer per-batch progress output.


steven-tey:
@coderabbitai full review

PR #4343, opened 2026-08-17
Feature/partner-search: write-path sync

Joker666:
# Partner search sync

Stacked on [#4333](https://github.com/dubinc/dub/pull/4333). Review that first. That PR made the index correct. This one keeps it correct and ships it.

Until now the only thing writing to the index was the backfill script, so the moment a partner was renamed, moved, banned, tagged, or given a link, it was stale with nothing to fix it.

## How it works

```
mutation ──▶ queuePartnerSearchSync({ ids }) ──▶ QStash ──▶ job re-reads the DB
                                                              │
                                            ┌─────────────────┴─────────────────┐
                                            ▼                                   ▼
                                    row found → upsert                row gone → delete
```

**No call site asks for a delete.** All 56 queue IDs. The job reads them back and decides. That falls out of reading state at execution time rather than trusting a dispatch-time snapshot, which also makes the job idempotent under retries.

## What changed

- **Sync layer**: `search/sync.ts`, `partner-search-sync-job` (self-continuing on the partner fan-out), and `queue-partner-search-sync.ts` (chunking and delay policy).
- **59 call sites**: status (16), deletions (11), links (10), profile (6), importers (6), platforms (4), creation (3), tags (2), group moves (1). Payout state, Veriff, discount and reward writers are deliberately not hooked, because they touch nothing the document carries.

<details>
<summary><b>The 59 call sites, grouped</b></summary>

**Enrollment status (16)** — approve, reject, invite, archive, ban, reactivate, deactivate
`lib/actions/partners/accept-program-invite.ts` · `archive-partner.ts` · `ban-partner.ts` · `unban-partner.ts` · `reactivate-partner.ts` · `decline-program-invite.ts` · `create-program-application.ts` · `bulk-approve-partners.ts` · `bulk-archive-partners.ts` · `bulk-ban-partners.ts` · `bulk-reject-partner-applications.ts` · `lib/api/partners/applications/approve-partner.ts` · `reject-partner.ts` · `bulk-reactivate-partners.ts` · `process-partner-deactivation.ts` · `app/(ee)/api/cron/partners/auto-reject/route.ts`

**Enrollment created (3)** — queued after default links exist
`lib/api/partners/create-and-enroll-partner.ts` · `lib/actions/partners/bulk-invite-partners.ts` · `lib/partners/complete-program-applications.ts`

**Deletions (11)** — all capture IDs before the delete
[...]

PR #4346, opened 2026-08-17
Feature/partner-search: fix search box debounce and guard precedence

Joker666:
# Fix the search box debounce and guard precedence

Stacked on [#4343](https://github.com/dubinc/dub/pull/4343). Three pre-existing defects in `SearchBoxPersisted`, called out but deliberately left alone in [#4316](https://github.com/dubinc/dub/pull/4316) because they change behaviour for all 20 consumers.

## Guard precedence, in two places

`!==` binds tighter than `??`, so both guards read `x ?? ("" !== y)` rather than `(x ?? "") !== y`. That evaluates to the param itself whenever one is present, which is always truthy, so neither guard ever skipped.

The visible consequence is on mount:

```
?search=foo&sortBy=relevance    guard: "foo" ?? ("" !== "foo")  ->  "foo"  (fires)
                                fixed: ("foo") !== "foo"        ->  false  (skips)
```

Firing calls `queryParams` with `del: ["page", ...resetParamsOnChange]`. On the partners table that list is `["sortBy", "sortOrder"]`, so opening a shared relevance-sorted search link stripped the sort before the user touched anything and the results quietly fell back to `totalSaleAmount` order.

## Clear button never cancelled the debounce

Clearing notifies immediately, but a keystroke still waiting out the 500ms landed afterwards and restored the term that was just cleared. Now cancelled first.

## Notes

`#4316` described this as two bugs. The precedence one is in two places, and the consumer count there is 20 files rather than 63.

No tests. This repo has no DOM environment and no component tests, so adding that setup would be a larger and more debatable change than the fix. Happy to do it separately if wanted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed search clearing so pending input updates no longer restore the previous search term.
  * Corrected synchronization between search values and URL parameters, ensuring updates are reflected accurately.


steven-tey:
@coderabbitai full review

steven-tey (approved):

PR #4432, opened 2026-09-01
Feature/partner-search: strip the program domain from pasted links

Joker666:
## Summary

Strip the program domain from a pasted short link before the query reaches Turbopuffer.

A paste like `go.acme.com/partner` sends the tokens `go`, `acme`, `com` and `partner`. The index stores link keys, not domains, but the domain tokens still match partner company names, websites and descriptions in the identity branch. Those matches widen the result set and can outrank the true owner of the key. This holds after the keys-only backfill too, so it is a separate fix from the re-index.

## Changes

- `stripProgramDomain`: reduces a link on the program's own domain to its key. Matches with or without protocol, `www.` and trailing slash, case-insensitive. Drops query strings and fragments. Links on other domains, the bare domain and plain queries are unchanged.
- `isLinkShapedQuery`: gates the program lookup so it only runs for link-shaped queries. Plain words and emails never trigger it.
- `resolvePartnerSearchCandidateQuery`: builds the candidate query and applies the strip. `getPartners` and `getPartnersCount` both use it, so the list and the count send the same query.

## Test plan

- [x] `partner-search-query.test.ts`: strip and shape cases (protocol, `www.`, case, trailing slash, query string, fragment, nested path, other domain, bare domain, email, no program domain).
- [x] `partner-search-resolve-query.test.ts`: program lookup only for link-shaped queries, strip applied, other domains untouched, null passthrough.
- [x] Existing `getPartners` and `getPartnersCount` search tests unchanged and passing.

## Summary by CodeRabbit

* **New Features**
  * Improved partner search for link-shaped queries, including URLs with protocols, paths, query parameters, and fragments.
  * Matching program domains are removed so searches use the relevant link path or key.
  * Partner lists and result counts now apply consistent search behavior.

* **Bug Fixes**
  * Links from other domains, invalid inputs, emails, and plain-text queries remain unchanged.

### DIFF
--- a/apps/web/lib/api/partners/get-partners-count.ts
+++ b/apps/web/lib/api/partners/get-partners-count.ts
@@ -10,10 +10,10 @@ import {
   mergePartnerCountryAndSearchWhere,
 } from "./program-enrollment-query";
 import {
-  buildPartnerSearchCandidateQuery,
   findPartnerSearchCandidates,
   getPartnerSearchReadProvider,
   PartnerSearchProvider,
+  resolvePartnerSearchCandidateQuery,
 } from "./search";
 
 type PartnersCountFilters = z.infer<typeof partnersCountQuerySchema> & {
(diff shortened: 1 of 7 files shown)


Joker666:
Recreated from the upstream branch so previews run: see the new PR.

PR #4433, opened 2026-09-01
Feature/partner-search: strip the program domain from pasted links

Joker666:
## Summary

When a user pastes a short link such as `go.acme.com/partner`, partner search sends both the domain and key as search tokens. The domain tokens can add unrelated matches and change the result order.

This change removes the program domain from matching short links and searches with the link key only. It does not require a search re-index.

## Changes

- Detect link-shaped search values.
- Remove the exact program domain from matching short links.
- Support protocols, `www.`, trailing slashes, query strings, fragments, and nested paths.
- Keep other domains, bare domains, plain text, emails, and partner IDs unchanged.
- Use the same resolved query for partner lists and counts.

## Test plan

- Added tests for link detection and domain removal.
- Added tests for program-domain lookup and unchanged queries.
- Confirmed that the existing partner list and count search tests pass.
- Added tests that the partner list and count send the stripped query to the provider.

## Validation against the production index

Ran `stripProgramDomain` on each paste with the Acme program domain, then sent the raw and the stripped query to the production Turbopuffer namespace.

| Paste | Query sent | Candidates | Top hit |
|---|---|---|---|
| `getacme.link/text-pplx` | `text-pplx` | 49 to 1 | stale e2e doc to Steven Elegance |
| `https://getacme.link/text-pplx` | `text-pplx` | 56 to 1 | same |
| `https://getacme.link/text-pplx/` | `text-pplx` | 56 to 1 | same |
| `getacme.link/e2-e-test-partner-goal-3cpw` | `e2-e-test-partner-goal-3cpw` | 729 to 729 | owner, both ways |
| `dub.sh/text-pplx` | unchanged | 741 to 741 | unchanged |

The single hit for `text-pplx` is the enrollment whose document carries that key. The e2e row shows the owner still ranks first, but the count stays at 729 with or without the domain. The last row is another domain, left untouched by design.

## Summary by CodeRabbit

* **New Features**
  * Partner search now recognizes program links and searches using the relevant link key.
  * Links with protocols, `www` prefixes, paths, query parameters, or fragments are handled consistently.
  * Search results and partner counts now apply the same link-processing behavior.

* **Bug Fixes**
  * Links hosted on other domains, plain text queries, emails, and partner IDs continue to search unchanged.
  * Improved handling of empty or invalid search input.

### DIFF
--- a/apps/web/lib/api/partners/get-partners-count.ts
+++ b/apps/web/lib/api/partners/get-partners-count.ts
@@ -10,10 +10,10 @@ import {
   mergePartnerCountryAndSearchWhere,
 } from "./program-enrollment-query";
 import {
-  buildPartnerSearchCandidateQuery,
   findPartnerSearchCandidates,
   getPartnerSearchReadProvider,
   PartnerSearchProvider,
+  resolvePartnerSearchCandidateQuery,
 } from "./search";
 
 type PartnersCountFilters = z.infer<typeof partnersCountQuerySchema> & {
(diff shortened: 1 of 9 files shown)


steven-tey (approved):

steven-tey:
@coderabbitai final review

PR #4474, opened 2026-09-09
Hide members from `viewer`, add `tenantId` to searchable fields

steven-tey:
## Summary by CodeRabbit

- **Access Control**
  - Workspace members and invitations are accessible to workspace owners, members, and billing roles.
  - Viewer-role users can no longer access the workspace members page.
  - Viewers see an access-denied message with a link to workspace settings.
  - Viewer accounts no longer load workspace member or invitation data when access is unavailable.

PR #4484, opened 2026-09-09
Feature/partner-search: select partner tags from the enrollment

Joker666:
## Summary

`partnerSearchDocumentSelect` read `programPartnerTags` through `partner`, which returns the partner's tags across every program, and the serializer filtered them back to the enrollment's program. `ProgramEnrollment` has its own `programPartnerTags` relation bound by `(programId, partnerId)`, so selecting there returns only this program's tags and the filter goes away.

The first version of the select did this (735d6515d3, Aug 8). The tags were dropped with the other filters (269e57820d, Aug 9) and re-added under `partner` (37faaf2d44, Aug 14). This restores the original shape.

## Changes

- Move `programPartnerTags` to the enrollment level, `partnerTagId` only.
- Drop the `programId` filter and the comment explaining it.
- Test fixtures build the source at the enrollment level. The cross-program leak case is gone, since the relation cannot return another program's tags.
- Fixtures gain `tenantId: null`, which #4474 added to the select without updating them.

## Test plan

- [x] `partner-search-document`, `partner-search-sync`, `partner-search-backfill`, `get-partners-search`, `turbopuffer-partner-search-provider`: 47 tests pass.
- [x] Typecheck clean on the touched files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved partner search serialization so partner tags are correctly associated with enrollment records.
  - Preserved accurate, deduplicated partner tag results in search documents.
  - Improved handling of search records without an associated tenant.

- **Tests**
  - Updated partner search backfill, document serialization, and synchronization coverage to reflect the current enrollment data structure.
  - Added coverage for enrollment records without a tenant and simplified partner tag data.

### DIFF
--- a/apps/web/lib/api/partners/search/serialize-document.ts
+++ b/apps/web/lib/api/partners/search/serialize-document.ts
@@ -1,5 +1,5 @@
-import { Prisma } from "@prisma/client";
 import { unique } from "@dub/utils";
+import { Prisma } from "@prisma/client";
 import { PartnerSearchDocument } from "./types";
 
 export const partnerSearchDocumentSelect = {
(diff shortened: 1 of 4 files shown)


steven-tey (approved):

steven-tey:
@coderabbitai full review

